PR #65, opened 2021-11-15
PROTOTYPE: Demonstrate module linking in Javy

smacleodify:
Javy now generates (and runs) wasm modules which are intended to be
linked with a JS engine module at runtime. The setup is pretty hacky
since Javy will create the module and run it everytime it is called,
but it's good enough to demonstrate how each piece should work.


freynca:
cc @mkcny @maximebedard @dwdking @DuoSRX @SpencerHaan

PR #76, opened 2022-02-14
Rename quickjs-sys crate to quickjs-wasm-sys

jeffcharles:
There's already a crate named quickjs-sys on crates.io so we need to
rename ours to be able to publish it.

Tested the changes with `make` and `make tests`.

I was not able to test the changes to the 3 quickjs benchmark crates as they try to pull the `quickjs-sys` crate from Github.

Could use pointers on anything I may have missed.


saulecabrera (approved):
I thought we had a step in CI to check benchmarks. We should probably add one after merging this one

PR #78, opened 2022-02-17
Make WASI STDIO the default io method for modules.

smacleodify:
Initial draft of using WASI STDIO by default.

The benchmarks are currently broken as they rely on the old API. Next steps to fix them would be:

1. Update Wasmtime version
2. Change the runner API to set STDIN on the WasiCtx
3. Figure out how to make the benchmarks not recreate the WasiCtx every run.

This will all need to change when module linking is introduced very soon, so it's probably best just to wait and fix them post module linking.


Makefile
```diff
@@ -29,31 +24,31 @@ test-cli: core
...
 
-tests: check-benchmarks test-core test-cli
+tests: test-core test-cli
 
 fmt: fmt-quickjs-wasm-sys fmt-core fmt-cli
 
 fmt-quickjs-wasm-sys:
 		cd crates/quickjs-wasm-sys/ \
-				&& cargo fmt -- --check \
+				&& cargo fmt \
```
jeffcharles, line 33: Can you provide context on removing the `-- --check`?

crates/core/src/engine.rs
```diff
@@ -1,72 +1,32 @@
...
+        Ok(output)
     }
-
-    let mut input_buffer = vec![0; len];
-    unsafe {
-        input_copy(input_buffer.as_mut_ptr());
+    #[cfg(feature = "json-io")]
+    {
+        let output: serde_json::Value = serde_json::from_slice(&output)?;
+        rmp_serde::to_vec(&output).map_err(Into::into)
```
jeffcharles, line 16: Would it be possible to transcode the JSON to msgpack?

crates/core/src/main.rs
```diff
@@ -43,16 +43,15 @@ pub extern "C" fn init() {
     }
 }
 
-#[export_name = "shopify_main"]
```
jeffcharles, line 46: I think I'm missing some context on this change

crates/core/src/engine.rs
jianghong: What's the value of doing that here?

jianghong:
I tested this with a JS script generated by our CLI tooling and verified this works on production.

crates/core/src/main.rs
jianghong: Oh the addition context here is that this `main` function by default is assigned the `_start` name

crates/cli/Cargo.toml
```diff
@@ -17,8 +17,9 @@ tempfile = "3.2.0"
 binaryen = "0.12.0"
 
 [dev-dependencies]
-wasmtime = "0.29"
-wasmtime-wasi = "0.29"
+wasmtime = "0.33"
+wasmtime-wasi = "0.33"
+wasi-common = "0.33"
```
saulecabrera, line 22: There's a security vulnerability in this version IIRC, we should be using `0.34.1`

saulecabrera (commented):
LGTM, once we update wasmtime to `v0.34.1`. 

This change will also require a new release: `v0.2.0`

crates/core/src/main.rs
jeffcharles: > Oh the addition context here is that this main function by default is assigned the _start name

Yeah that was the piece I was missing 🙂. Thank you for elaborating!

crates/cli/Cargo.toml
saulecabrera: Yeah, I'll submit an update to Wizer, we'll need to follow up on this here.

jeffcharles (approved):

crates/core/src/engine.rs
jianghong: Done here https://github.com/Shopify/javy/pull/78/commits/f3a9cf306329fc54657df1bd6579ce2328f37eb9

jianghong:
Addressed Jeff's comment of transcoding from json to msgpack directly in this commit https://github.com/Shopify/javy/pull/78/commits/f3a9cf306329fc54657df1bd6579ce2328f37eb9

jeffcharles (approved):

PR #83, opened 2022-02-24
Add installing wask-sdk to publish.yml step

jianghong:
publish action was failing because the environment needs to do install wasi-sdk first. 

https://github.com/Shopify/javy/actions/runs/1893609362

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,6 +19,9 @@ jobs:
           toolchain: 1.53.0
           target: wasm32-wasi
           default: true
+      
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,9 +1,7 @@
 name: Publish
 
 on:
-  release:
-    types:
-      - created
+  push:
```
jianghong, line 4: Will change this back after I test and verify this works

jeffcharles (approved):

PR #88, opened 2022-03-04
Send console.log and console.error to stderr

jianghong:
Fixes https://github.com/Shopify/javy/issues/87

- Send `console.log` calls to stderr so that it doesn't conflict with the output being sent to stdout 
- Implement `console.error` and send that to stderr

### DIFF
--- a/crates/cli/tests/runner/mod.rs
+++ b/crates/cli/tests/runner/mod.rs
@@ -102,7 +102,7 @@ impl Runner {
         let logs = store_context
             .log_stream
             .try_into_inner()
-            .expect("Output stream reference still exists")
+            .expect("log stream reference still exists")
             .into_inner();
         let output = store_context
             .wasi_output
--- a/crates/cli/tests/runner/mod.rs
+++ b/crates/cli/tests/runner/mod.rs
@@ -16,14 +16,21 @@ pub struct Runner {
 struct StoreContext {
     wasi_output: WritePipe<Cursor<Vec<u8>>>,
     wasi: WasiCtx,
+    log_stream: WritePipe<Cursor<Vec<u8>>>,
 }
 
 impl Default for StoreContext {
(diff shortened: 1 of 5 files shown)


crates/core/src/main.rs
```diff
@@ -23,7 +23,9 @@ static SCRIPT_NAME: &str = "script.js";
 pub extern "C" fn init() {
     unsafe {
         let mut context = Context::default();
-        context.register_globals(io::stdout()).unwrap();
+        context
+            .register_globals(io::stderr(), io::stderr())
```
jianghong, line 27: I'm sure there's a way to just pass this in once but I wouldn't figure out, happy to pair with someone on it if you know

crates/quickjs-wasm-rs/src/js_binding/context.rs
```diff
@@ -359,4 +361,21 @@ mod tests {
         assert_eq!(b"bonjour le monde\n", stream.0.borrow().as_slice());
         Ok(())
     }
+
+    #[test]
+    fn test_console_error() -> Result<()> {
```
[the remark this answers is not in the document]
jianghong: Fixed here https://github.com/Shopify/javy/pull/90

crates/cli/tests/runner/mod.rs
```diff
@@ -89,11 +99,17 @@ impl Runner {
         run.call(&mut store, ())?;
         let store_context = store.into_data();
         drop(store_context.wasi);
-        Ok(store_context
+        let logs = store_context
+            .log_stream
+            .try_into_inner()
+            .expect("Output stream reference still exists")
```
jeffcharles, line 105: Should probably update this text to indicate it's referring to the log stream or stderr and not the output

saulecabrera (approved):

PR #92, opened 2022-03-09
Compile modules for linking

jianghong:
Credit to @smacleodify and @saulecabrera for a big chunk of the exploration and figuring out a lot of this!

This PR changes javy completely to compile a JS file to a wasm module that imports and links against a JS runtime wasm module. The standalone JS wasm will not run unless properly linked against the JS runtime wasm.

### Core
This is where we build the JS runtime. The export API is defined to be

- init_engine
- init_src
- execute
- canonical_abi_realloc
- canonical_abi_free

This can be built with
```
make wizen-core
```

What this is doing under the hood is call wizer with the `init_engine` function which gives us a wasm with a memory snapshot that has the JS engine context all set up already. 

This module will be shipped to our runtime engine and be used to link against partner modules that import this module.

### Cli
We handcraft our own wasm file in `js_module.wat`. This is where all the js runtime imports are defined. This is also used as a template where we can inject the partner's JS source code 

There are two functions in this module

 - shopify_init 
 - _start

The CLI takes the input JS source file and puts the source directly into the `js_module.wat` in the data section. This module is then importing and using the core module API to run and execute the JS code

### Wizer points
Something that is important and interesting to discuss is different points that we want to wizer. This affected how we designed the core module API, as well as the JS module API

The first point is we call wizer with the core module and `init_engine` function. This gives us a wasm module with a memory snapshot of the JS engine globals already being set up.

The second point is we can further wizer the partner's JS code module on `shopify_init` function. This wizened module can be done ahead of time and the execution time will save on the the initializing steps in `shopify_init`. Note that this second wizen step is currently not possible https://github.com/bytecodealliance/wizer/blob/42e3948fef69e4eeb5de51eb8ec6786fbecd430c/src/parse.rs#L330-L333

But our API design sets us up for these future wizer optimizations.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "javy"
-version = "0.3.0"
+version = "0.4.0"
 authors = ["Saúl Cabrera <saulecabrera@gmail.com>"]
 edition = "2018"
 build = "build.rs"
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -1,8 +1,8 @@
 mod engine;
 
-use quickjs_wasm_rs::{messagepack, Context};
+use quickjs_wasm_rs::{json, Context};
 
-use messagepack::{transcode_input, transcode_output};
+use json::{transcode_input, transcode_output};
 use once_cell::sync::OnceCell;
 use std::alloc::{alloc, dealloc, Layout};
 use std::io::{self};
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -971,7 +971,7 @@ checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "javy"
-version = "0.3.0"
+version = "0.4.0"
 dependencies = [
  "anyhow",
  "binaryen",
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ For more information on the commands you can run `javy --help`
 
 ## Releasing
 
-1. Bump the version number in the Cargo.tomls that make sense
+1. Bump the version number in the cli/Cargo.toml
 2. Create a tag for the new version like `v0.2.0`
 ```
 git tag v0.2.0


crates/cli/Cargo.toml
```diff
@@ -1,20 +1,20 @@
 [package]
 name = "javy"
-version = "0.1.0"
+version = "0.3.0"
 authors = ["Saúl Cabrera <saulecabrera@gmail.com>"]
 edition = "2018"
-build = "build.rs"
+build = "src/build.rs"
```
saulecabrera, line 6: Why is  this location change needed?

saulecabrera: I think I understand; it seems that this change is needed because now the build process relies on the `Opt` module; with these new changes the opt-module becomes a build dependency, so instead of including the build script into the src, I'd prefer if we switch the relationship and move all the code from the Opt module inside the build script, keeping the the build script at the root. That will also make it consistent with the other newly-declared build dependencies (`wizer`, `binaryen`, etc)

saulecabrera (approved):
Looks good to me; I do have a small comment around the new location of the build script, I think that we can refactor that piece out to keep things standard.

crates/core/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+
+/// Executes the JS code.
+/// func_obj_path is expected to be a dot spearate path to the main function.
+///
+/// # Safety
+///
+/// TODO: When can this be called safely?
+#[export_name = "execute"]
+pub unsafe extern "C" fn execute(
+    func_obj_path_is_some: u32,
```
jeffcharles, line 59: I'm out of the loop on this, what is this parameter supposed to represent? I can see how it's used below to say use the path pointer and length if this is non-zero and use `"Shopify.main"` otherwise. Is this parameter something proprietary to Shopify or something more standard?

jeffcharles (approved):
Changes look good to me, just have some nits and questions. Could also be good to get some of the safety todo's filled in.

crates/core/src/lib.rs
jeffcharles: If this is Shopify specific, I guess as a follow-up nit, can we change this to `func_obj_path_is_defined` or `func_obj_path_is_provided`? I get where `is_some` is coming from but it initially confused me.

smacleodify: This is how `option<string>` in `.wit` is handled by the canonical ABI. It is translated to `(u32, u32, u32)` where the first argument indicates if the argument is present or not. `..._is_some` is used for consistency with rusts `Option<...>` type.

This argument suffix we use is never exposed to actual consumers of the library, they would see something in `.wit` like `execute: function(func-obj-path: option<string>)` and the actual webassembly file does not have named arguments.

jianghong:
We have decided to not ship this for now. When we come back to this, the implementation will likely have to change so closing this for now

PR #97, opened 2022-03-29
msgpack -> json

jianghong:
As part of https://github.com/Shopify/script-service/issues/4539

Assume input and output as json and not messagepack

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,12 +165,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "byteorder"
-version = "1.4.3"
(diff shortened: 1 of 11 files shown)


jeffcharles:
IMO `json` should be a feature of the `quickjs-wasm-rs` crate, not something that's compiled in unconditionally. The main reason is other users of `quickjs-wasm-rs` like Suborbital may want to use a different serialization mechanism (I think Suborbital uses a wit-bindgen approach) and including json parsing the in the compiled module unconditionally increases the size of the module unnecessarily for those users.

PR #100, opened 2022-04-04
Updated main heading in README.md

Fienne:
Fixes https://github.com/Shopify/javy/issues/99


README.md
```diff
@@ -1,4 +1,4 @@
-# Javy: A _Jav_ aScript to WebAssembl _y_  toolchain
+# Javy: A _JavaScript to WebAssembly_  toolchain
```
saulecabrera, line 1:
```suggestion
# Javy: A *Jav*aScript to WebAssembl*y* toolchain
```

It seems that using `*` rather than `_` does what we want

PR #101, opened 2022-04-06
Run release workflow on `published`

jbourassa:
I was confused as to why the GitHub workflow wasn't triggered on the
Javy 0.3.0 release, so I figured I'd try to make things a little better.

With this, the publish workflow will (should?) run when a draft
release is published.

I've also updated the README to mention that.

---

The PR has 2 commits. The first one is only linter fixes to `publish.yml`,
feel free to gloss over it.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,7 +3,7 @@ name: Publish
 on:
   release:
     types:
-      - created
+      - published
 
 jobs:
   compile_core:
--- a/README.md
+++ b/README.md
@@ -53,12 +53,12 @@ For more information on the commands you can run `javy --help`
 
 1. Create a tag for the new version like `v0.2.0`
 ```
-git tag tag v0.2.0
+git tag v0.2.0
 git push origin --tags
 ```
-2. Create a new release from the new tag in github [here](https://github.com/Shopify/javy/releases/new)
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -4,22 +4,23 @@ on:
   release:
     types:
       - created
+      - published
```
jeffcharles, line 7: Do we want both `created` and `published` or just `published`? As far as I can tell, `created` will cause a run when a `draft` is saved. Do we want that behaviour?

jeffcharles (approved):
I have the question in there but I can't think of a reason to not leave `created` in there.

.github/workflows/publish.yml
jeffcharles: > I put in both in there because I wasn't sure that a new published release would trigger the published event

Based on my reading of the docs, it should run in that case.

From https://docs.github.com/en/github-ae@latest/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-32:
> `published`: a release, pre-release, or draft of a release is published

A new published release _should_ fall into the set of events that fall under a release that's published if I'm interpreting that right.

jeffcharles (approved):

PR #106, opened 2022-05-17
Elaborate on compilation instructions

jeffcharles:
Adds an example of what the source JavaScript looks like and how to interact with the input as well as how to write the output.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,13 @@ jobs:
             ${{ runner.os }}-cargo-target
 
       - name: Install wasmtime-cli
-        run: cargo install wasmtime-cli
+        env:
+          WASMTIME_VERSION: 0.34.1
+        run: |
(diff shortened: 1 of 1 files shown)


saulecabrera (approved):

saulecabrera:
Oh it seems that the failure is because wasmtime is requiring a different rust version (since I'm building it from source). The process of building it from source is very slow, we should probably change it to installing the pre-built binaries like I did here: https://github.com/bytecodealliance/spidermonkey-wasm-rs/pull/36/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

We could: (i) merge as is, since this PR is not changing anything code related, we'll need to get to it eventually though (ii) define the rust version required by wasmtime (iii) add the pre-build binaries. I'll leave it up to you

jeffcharles:
I'll do that as a separate PR and then rebase this on top of that one once it's merged

PR #112, opened 2022-05-26
[ci skip] docs: Explicitly document the WASI-only nature of Javy modules

saulecabrera:
Follow up to https://github.com/Shopify/javy/issues/111

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,14 @@ $ echo '{ "n": 2, "bar": "baz" }' | wasmtime index.wasm
 {"foo":3,"new_bar":"baz!"}%   
 ```
 
+### Invoking Javy-generated modules programatically
+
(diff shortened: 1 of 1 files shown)


codefromthecrypt (commented):
drive by, but you could clarify this is I believe about the "WASI command module" pattern, and specifically the "snapshot-01" version. I believe the group is doing another incompatible WASI version as snapshot-02 now or soon.

https://github.com/WebAssembly/WASI/blob/snapshot-01/design/application-abi.md

jeffcharles (approved):

codefromthecrypt:
ps I'm wondering if use of os args (args_get args_sizes_get) shouldn't be called out explicitly? I suspect for some use cases, this would be easier than stdin..

PR #122, opened 2022-08-09
Allow different Write types to be registered as console globals

nico151999:


saulecabrera:
The change looks good to me. Do you mind signing the CLA?

saulecabrera:
Yeah I had to re-run the job. Thanks!

PR #124, opened 2022-08-10
Update wizer 

saulecabrera:
I believe this will fix the security vulnerability notifications.


saulecabrera:
All of the issues can be resolved by using `wasmtime-0.34.2`, and we already have that, wizer gives us an indirect dependency to wasmtime; the latest main points to 0.34.2

jeffcharles (approved):

PR #129, opened 2022-10-13
Introduce WASI global 

surma:


jeffcharles:
Is the idea to have this in addition to `console.log` and `console.error`?

surma:
It’s totally still up for bikeshedding, but yes, my idea was that all low-level WASI imports that we want to expose, go onto the `WASI` global. _In addition_ we should also expose them where ever is idiomatic (so `console.log` for stdout, `Math.rand()` for random numbers, once WASI has that etc)

PR #137, opened 2022-11-07
Fix rerun condition

Gentle:
the current rerun conditions for quickjs-wasm-sys builds include a non-existing path

Cargo's behavior is to rebuild every time if any of the files is missing, so that was the current behavior

I fixed the wrong rerun-if-changed condition and added another one to make sure the crate gets rebuilt any time the C code was changed


jeffcharles:
Closing since this has been addressed in https://github.com/Shopify/javy/pull/174 and https://github.com/Shopify/javy/pull/168.

PR #141, opened 2022-11-21
Modifying the sanitization of json keys to use camel instead of snake casing

dwdking:
#gsd:31055

Modifying the sanitization of json keys to use came instead of snake casing

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -56,7 +56,7 @@ fn main() -> Result<()> {
         ),
     )?;
 
-    memory.write(&mut store, contents_ptr.try_into()?, &mut buffer)?;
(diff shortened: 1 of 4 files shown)


dwdking:
> Should we also remove the camel-casing process on serialization?

@saulecabrera, I think the plan is to remove the automatic serialization/deserialization so I don't know if it would really matter. I didn't have to modify the serialization in order to get the function to work so it's not required at this point. I also don't have a good understanding of everything the `sanitize_key` method is trying to accomplish.

jeffcharles:
I just merged https://github.com/Shopify/javy/pull/148 which has Javy Core no longer perform the JSON serialization and relies entirely on the provided JS to do it. Do we still need this PR with that in?

PR #143, opened 2022-11-28
make key case conversion optional in `serialize` module

dicej:
This adds new constructor functions for `Serializer` and `Deserializer` which disable key case conversion.  This is useful for when we want keys to remain unchanged during conversion to and from JS objects.

Signed-off-by: Joel Dice <joel.dice@fermyon.com>

### DIFF
--- a/crates/quickjs-wasm-rs/src/serialize/de.rs
+++ b/crates/quickjs-wasm-rs/src/serialize/de.rs
@@ -16,6 +16,7 @@ impl SerError for Error {
 pub struct Deserializer {
     value: Value,
     map_key: bool,
+    convert_key_case: bool,
 }
 
 impl From<Value> for Deserializer {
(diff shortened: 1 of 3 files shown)


saulecabrera:
Hey @dicej, I assume that this is useful for your use-case? If so, we are planning on removing the serialization pass and camel-casing as a whole. Would that change work for you?

dicej:
Yes, I'd love to see the case conversion (in both the `serialize` and `deserialize` directions) removed completely -- it's only been causing issues for us.

dicej:
Thanks for the update @jeffcharles.  We're not using the `cli` or `core` crates -- just `quickjs-wasm-js` and `quickjs-wasm-sys`, so #148 doesn't really help us.  So yes, we'd like to remove the case conversion from the serializer and deserializer if the consensus is that it's no longer needed.  Happy to close this PR and open a new one to that effect if it's helpful.

PR #145, opened 2022-12-01
Add TextEncoder and TextDecoder APIs

jeffcharles:
First part of a set of changes to alter how JS code is executed inside Javy. Also adds a JS script that is evaluated before the users' JS to have an easier time adding new JS APIs that would be more difficult to define in Rust.

### DIFF
--- a/crates/quickjs-wasm-rs/src/js_binding/context.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/context.rs
@@ -1,7 +1,7 @@
 use super::constants::{MAX_SAFE_INTEGER, MIN_SAFE_INTEGER};
 use super::exception::Exception;
 use super::value::Value;
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use once_cell::sync::Lazy;
 use quickjs_wasm_sys::{
     ext_js_exception, ext_js_null, ext_js_undefined, size_t as JS_size_t, JSCFunctionData,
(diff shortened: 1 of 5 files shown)


crates/quickjs-wasm-rs/src/js_binding/context.rs
```diff
@@ -310,6 +324,56 @@ where
...
+        let view = buffer
+            .get(byte_offset..(byte_offset + byte_length))
+            .ok_or(anyhow!(
+                "Provided offset and length is not valid for provided buffer"
+            ))?;
+        let str = if fatal {
+            Cow::from(str::from_utf8(view).map_err(|_| anyhow!("The encoded data was not valid"))?)
+        } else {
+            String::from_utf8_lossy(view)
+        };
```
saulecabrera, line 361: I'm a bit confused here, the spec states:

> If the property is true then a decoder will throw a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) if it encounters malformed data while decoding. If false the decoder will substitute the invalid data with the replacement character U+FFFD (�). The value of the property is set in the [TextDecoder() constructor](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder).

According to this description, shouldn't we be doing something like:

```rust
let err = str::from_utf8(view).map_err(...);

if result.is_ok() {
   // return 
}

if fatal {
   // ... type error
} else {
   // ... substitute the view with the right character
}
```

saulecabrera: You're right, my bad.

PR #147, opened 2022-12-02
Use Cargo inheritance and cleanup Cargo.tomls

saulecabrera:
This commit uses cargo inheritance to declare the common dependendencies in the root Cargo.toml and also cleans up the Cargo.toml definitions.


jeffcharles (approved):
🥇

PR #148, opened 2022-12-02
Stop JSON transcoding input and output

jeffcharles:
Transcoding JSON input and output no longer offers a performance benefit over JS code performing serialization and makes some use cases for Javy more difficult or impossible.


crates/cli/tests/integration_test.rs
```diff
@@ -44,70 +42,49 @@ fn test_str() {
...
 
-fn run<I, O>(r: &mut Runner, i: &I) -> (O, String)
-where
-    I: Serialize,
-    O: DeserializeOwned,
-{
-    let input = serde_json::to_vec(i).unwrap();
-    let (output, logs) = r.exec(input).unwrap();
-    let output = serde_json::from_slice::<O>(&output).unwrap();
+fn run_with_u8s(r: &mut Runner, i: u8) -> (u8, String) {
```
saulecabrera, line 80: We should just rename this to `run`, the signature is self descriptive.

saulecabrera (approved):
LGTM, just a small nit.

PR #150, opened 2022-12-08
remove case conversion from `serialize` module

dicej:
Per the discussion in #143, we've decided case conversion that was being done in `sanitize_key` is no longer desirable.  This commit removes it.

Signed-off-by: Joel Dice <joel.dice@fermyon.com>


jeffcharles (approved):
Thank you!

PR #158, opened 2022-12-13
Update benchmarks to use new contract

jeffcharles:


saulecabrera (commented):
Can you document here (in the PR) or in the README or in a dedicated README inside the benchmarks how to update this code in case the execution mechanism changes? Arguably, it'll be better if we had a high-level definition of this code such that we can transpile instead of having to update minified code. Do we have the original source of this?

saulecabrera (approved):
Discussed this offline; we'll come up with a better approach to benchmarks after landing this change.

PR #163, opened 2022-12-15
Move injected globals into core crate

jeffcharles:
Moves `console_log_to`, `decode_utf8_buffer_to_js_string`, and `encode_js_string_to_utf8_buffer` into the `core` crate. Moves the existing globals in core into a `globals` module since the `main` module was getting large. Also consolidated from functions injecting globals down to one function. Also removes the `engine` module from the `core` crate since we're no longer using anything in the `engine` module.

If this is too large, I can break into smaller PRs.

The one tradeoff is because the `console_log_to` function uses raw QuickJS types, I had to add a dependency from Core to `quickjs-wasm-sys`. I'd like to fix that in a future PR.

### DIFF
--- a/crates/core/src/globals.rs
+++ b/crates/core/src/globals.rs
@@ -1,9 +1,9 @@
 use anyhow::anyhow;
-use quickjs_wasm_rs::{Context, Value};
-use quickjs_wasm_sys::{
-    ext_js_exception, ext_js_undefined, size_t as JS_size_t, JSContext, JSValue, JS_FreeCString,
-    JS_ToCStringLen2,
+use quickjs_wasm_rs::sys::{
+    ext_js_exception, ext_js_undefined, JSContext, JSValue, JS_FreeCString, JS_ToCStringLen2,
+    JS_size_t,
 };
+use quickjs_wasm_rs::{Context, Value};
 use std::borrow::Cow;
 use std::ffi::c_int;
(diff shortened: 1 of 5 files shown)


crates/core/src/globals.rs
```diff
@@ -0,0 +1,249 @@
+use anyhow::anyhow;
+use quickjs_wasm_rs::{Context, Value};
+use quickjs_wasm_sys::{
+    ext_js_exception, ext_js_undefined, size_t as JS_size_t, JSContext, JSValue, JS_FreeCString,
+    JS_ToCStringLen2,
+};
```
saulecabrera, line 6: Can we add a `TODO` or `FIXME` to indicate that this patter should be fixed and should not be adopted?

saulecabrera: Alternatively, one "easier" fix would be to re-export the sys types from the core crate. I think that generally I prefer that as a short term measure.

surma (approved):

crates/core/src/globals.rs
jeffcharles: So I added the deprecated attribute but I couldn't get the warnings to actually show up. But I don't to spend too much time looking into that if I could spend the equivalent time on rewriting that code to use `wrap_callback` instead so we can get rid of the re-export entirely.

saulecabrera (approved):

PR #166, opened 2022-12-16
Add tests for javy lib

surma:
(These have _not_ been integrated with CI yet. Wanted to get your review first.)

This PR adds a bit of a test harness to run snippets of JS code through a bundler (to pull in our library) and then execute it with Javy with proper streaming I/O so we can test all the edge cases. Added two basic tests for now, but we can grow that.


npm/javy/tests/runner.js
```diff
@@ -0,0 +1,161 @@
...
+
+import * as tests from "./tests.js";
+
+const javyPath = new URL("../../../target/release/javy", import.meta.url)
+	.pathname;
+
+async function main() {
+	console.log("Running tests...");
+	const resultPromises = Object.entries(tests).map(([testName, testFunc]) =>
+		Promise.resolve(testFunc())
```
jeffcharles, line 25: Nit: Any reason not to use `await` on `testFunc()`?

npm/javy/tests/tests.js
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export async function longDelay() {
+	await runJS({
+		source: "./fixtures/lib_echo.js",
+		expectedOutput: "1234",
+		stdin: customStringInputStream(async controller => {
+			for (let i = 1; i <= 4; i++) {
+				controller.enqueue(i.toString());
+				await sleep(100);
```
jeffcharles, line 27: Is there a different way we could try to exercise the code path that the sleep is exercising but without sleeping? I would prefer we don't introduce sleeps into the test suite. (I know CI already takes forever to run and an extra 400 ms isn't making it materially worse but I'd like to know if there's a different way forward)

jeffcharles (commented):
Mostly looks good! Just have the one question.

npm/javy/tests/runner.js
surma: Instead of this, I could  do a `try/catch` + `await` (because I need to catch when `testFunc` throws). I found this more obvious, but I don't have a strong feeling.

jeffcharles: My personal style preference is to use `await` with a `try/catch` so the code can be read as if it were run sequentially. But it's not a super strong preference.

npm/javy/tests/tests.js
jeffcharles: I'd be with passing a `0` instead of `100`. That way it's not synchronous but it's also the shortest delay possible. If there's a concern that the code may assume that all input is available synchronously, I think that should flush that error case out.

npm/javy/README.md
```diff
@@ -0,0 +1,9 @@
+# Javy Lib
+
+This library provides abstractions and convenience methods over Javy's built-in functions.
+
+More documentation TBD>
```
saulecabrera, line 5: Are we missing an opening `<` here? Should we just remove this sentence altogether?

npm/javy/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Tests
+
+To run the tests, run `npm test`. It requires you to have built `javy` and have `wasmtime` installed.
```
saulecabrera, line 9: Maybe it's worth specifying that we should build Javy in release mode, since the release path is the one that is hardcoded.

saulecabrera (approved):
LGTM, just two small comments.

jeffcharles (approved):

PR #169, opened 2022-12-16
Use wrap_callback for console_log_to

jeffcharles:
Switch `console_log_to` to use `wrap_callback` instead of `new_callback`. This lets us get rid of the `sys` re-exports from `quickjs-wasm-rs`.

I also added a test case for passing non-string values to `console.log` to check if they're transformed as expected.


saulecabrera (approved):
Yeah, exactly what I was hoping for!

PR #170, opened 2022-12-16
Use crates.io instead of Git for Wizer

jeffcharles:
Looks like a recent version of Wizer is available on crates.io. Was hoping switching to that would let us update to a version of Wasmtime that doesn't have a security warning but Wizer's main branch still has a dependency on `0.35.3` of a shared dependency so no luck with that.

What this change would allow us to do is publish the CLI crate to crates.io. Also it's probably easier to understand the implications of version bumps if we use crate versions instead of Git sha's for tracking changes to Wizer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,18 +387,18 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.81.2"
+version = "0.82.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0eba0f73ab0da95f5d3bd5161da14edc586a88aeae1d09e4a0924f7a141a0093"
+checksum = "38faa2a16616c8e78a18d37b4726b98bfd2de192f2fdc8a39ddf568a408a0f75"
 dependencies = [
  "cranelift-entity",
 ]
 
 [[package]]
 name = "cranelift-codegen"
-version = "0.81.2"
+version = "0.82.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9cff8758662518d743460f32c3ca6f32d726070af612c19ba92d01ea727e6d9"
+checksum = "26f192472a3ba23860afd07d2b0217dc628f21fcc72617aa1336d98e1671f33b"
 dependencies = [
  "cranelift-bforest",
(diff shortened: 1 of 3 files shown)


saulecabrera:
The reason why we've been keeping the git dependency is that it allows us to have access to Wizer with the latest Wasmtime faster. Can we perhaps update Wasmtime in Wizer and request a new version? FWIW[ I've done that in the past](https://github.com/bytecodealliance/wizer/pulls?q=is%3Apr+author%3Asaulecabrera+is%3Aclosed). 


> What this change would allow us to do is publish the CLI crate to crates.io
Des this mean that you cannot publish to crates.io if your crate has a git dependency?

jeffcharles:
> The reason why we've been keeping the git dependency is that it allows us to have access to Wizer with the latest Wasmtime faster.

Understandable but I think using a crate version and trying to encourage the maintainers to cut a new release and only using the Git dependency when we have to seems like it would have maintainability benefits for us.

> Can we perhaps update Wasmtime in Wizer and request a new version?

Yeah I can do that.

> Des this mean that you cannot publish to crates.io if your crate has a git dependency?

Yep. From https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories:
> Note: [crates.io](https://crates.io/) does not allow packages to be published with git dependencies (git [dev-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) are ignored). See the [Multiple locations](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) section for a fallback alternative.

saulecabrera:
> Understandable but I think using a crate version and trying to encourage the maintainers to cut a new release and only using the Git dependency when we have to seems like it would have maintainability benefits for us.

I think it's ideal, but sometimes I'd prefer to use a git version _especially_ if there are security vulnerabilities in the middle or features from Wasmtime that we need.

jeffcharles:
Oh for sure, temporarily switching back to a Git version in the event of a security vulnerability makes sense. But, in the event of a security vulnerability, I think it should be using a Git dependency temporarily with the aim of getting back on a versioned crate when a version is released that removes the security vulnerability.

PR #180, opened 2023-01-06
Fix encoding of unpaired surrogates and enable WPT tests in CI

jeffcharles:
Fixes failing WPT `textencoder-utf16-surrogates` tests and adds any tests we've enabled to run in CI. We replace unmatched surrogates in JS with a UTF-16 replacement char before converting to UTF-8. This does introduce a 0.0862% to 3.5610% increase in execution time as measured by our benchmarks for cases where there are no unpaired surrogates.

The code is not the cleanest JS I could have written. I tried with a cleaner looking version and I saw 10% increases in total execution time. After some experimenting and trying to optimize for cases without surrogates, I got it down to what's here. There might be some other ways to get the time spent in this function down even further.

Some other approaches I considered:
- Addressing this in Rust code but we do not have access to the underlying WTF-16 byte array in the `JSValue` we receive as an argument. QuickJS does not expose any APIs to retrieve these bytes either. So we cannot preprocess the input in Rust without adding extensions to our QuickJS implementation.
- Briefly looked at if we could post-process the output we receive back from QuickJS from invoking `JS_ToCStringLen2` and after experimenting a bit, right now I'm not confident that there's a contiguous range of output byte triples where it would be safe to replace them with a UTF-8 replacement char without unintentionally replacing valid byte triples. It may be worth that exploring that further though. 
- Adding an extension to our QuickJS code to handle the string conversion better but, I am rejecting that approach for now given the approach here is more maintainable and the impact on execution speed seems manageable to me. We could do this in the future if it makes sense to.

For cases where there are unpaired surrogates, our current encoder implementation will currently throw an exception, so the slower execution speed that would result from this change is likely better in most cases with unpaired surrogates.

### DIFF
--- a/crates/core/prelude/text-encoding.js
+++ b/crates/core/prelude/text-encoding.js
@@ -47,8 +47,9 @@
             });
         }
 
-        encode(input) {
+        encode(input = "") {
             input = input.toString(); // non-string inputs are converted to strings
+            input = replaceUnpairedSurrogates(input); // QuickJS does not handle this correctly
             return new Uint8Array(__javy_encodeStringToUtf8Buffer(input));
         }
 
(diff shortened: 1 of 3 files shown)


crates/core/prelude/text-encoding.js
```diff
@@ -47,8 +47,9 @@
             });
         }
 
-        encode(input) {
+        encode(input = "") {
             input = input.toString(); // non-string inputs are converted to strings
+            input = replaceUnpairedSurrogates(input); // QuickJS does not handle this correctly
```
surma, line 52: Sorry to question this, but are you sure about this? QuickJS passes the ECMA262 test suite, so I'd be surprised if there was any non-standard behavior in QuickJS. 

Isn't this exactly what WTF-16 allows, i.e. lone surrogates?

saulecabrera (commented):
> This does introduce a 0.0862% to 3.5610% increase in execution time as measured by our benchmarks for cases where there are no unpaired surrogates.

This is a significant bump in execution time and this is one of the things that worries me the most regarding our JS prelude; generally I'd like to minimize the amount of JS that we introduce as part of the prelude, especially when the side effects are as significant as this one.

Personally I feel that the increase in runtime performance is reason enough to dive deeper into alternatives 2 and 3.

jeffcharles:
I'm going to close this since I think I can use a different approach after reviewing the WTF-8 spec and my own notes on various outputs I was getting back from `JS_ToCStringLen2`.

PR #181, opened 2023-01-06
Enables one suite and adds fixmes for other relevant ones

jeffcharles:
- Enables the API basics tests for encoding
- Adds other test suites we should enable at some point with a comment indicating why they're disabled
- Copies an existing suite that includes UTF-16 inside the test case and comments out running the UTF-16 pieces
- Adds a safe default for `.encode`'s `input` argument

I want something in the code to be able to track which suites we want to turn on. These changes don't include suites we will never turn on like ones for legacy encodings. We can tackle the script importing not working as expected separately. It's mostly trying to pull in a list of valid encodings and we only care about UTF-8 so we can maybe do something else to make those tests happy or perhaps just create custom copies of them.

### DIFF
--- a/crates/core/prelude/text-encoding.js
+++ b/crates/core/prelude/text-encoding.js
@@ -47,7 +47,7 @@
             });
         }
 
-        encode(input) {
+        encode(input = "") {
             input = input.toString(); // non-string inputs are converted to strings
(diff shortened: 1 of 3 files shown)


surma (approved):
LGTM. I'll take a look at the script importing issue

PR #185, opened 2023-01-10
Fix bug with encoding unpaired surrogates

jeffcharles:
Fix failing WPT cases for text encoding containing unpaired surrogates. Benchmarks show no change in performance. Uses similar logic to `::from_utf8_lossy` but treats lone surrogates as having an error length of 3 instead of a separate error per byte so the correct number of replacement chars are inserted.

### DIFF
--- a/crates/quickjs-wasm-rs/src/js_binding/value.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/value.rs
@@ -174,12 +174,15 @@ impl Value {
                     // see https://simonsapin.github.io/wtf-8/#surrogate-byte-sequence
                     let lone_surrogate =
                         matches!(after_valid, [0xED, 0xA0..=0xBF, 0x80..=0xBF, ..]);
-                    // Rust's `Utf8Error` reports a lone surrogate as having an error_len of 1
-                    // which results in 3 replacement chars being inserted instead of 1. 1
-                    // replacement char is correct according to
-                    // https://simonsapin.github.io/wtf-8/#converting-wtf-8-utf-8. So, substitute an
-                    // error_len of 3 for lone surrogates so the surrogate is only counted once
-                    // instead of as 3 separate characters.
+
+                    // https://simonsapin.github.io/wtf-8/#converting-wtf-8-utf-8 states that each
+                    // 3-byte lone surrogate sequence should be replaced by 1 UTF-8 replacement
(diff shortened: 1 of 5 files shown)


surma:
Looked at the CI failure. `make tests` passes for me locally — so I wonder if something is going wrong with cloning the submodule?

crates/quickjs-wasm-rs/src/js_binding/value.rs
```diff
@@ -147,13 +149,60 @@ impl Value {
...
+    pub fn as_str_lossy(&self) -> std::borrow::Cow<str> {
+        let mut buffer = self.as_wtf8_str_buffer();
+        match str::from_utf8(buffer) {
+            Ok(valid) => Cow::Borrowed(valid),
+            Err(mut error) => {
+                let mut res = String::new();
+                loop {
+                    let (valid, after_valid) = buffer.split_at(error.valid_up_to());
+                    res.push_str(unsafe { str::from_utf8_unchecked(valid) });
+                    const REPLACEMENT_CHAR: char = '\u{FFFD}';
```
saulecabrera, line 171: I believe this can be imported from https://doc.rust-lang.org/stable/std/primitive.char.html#associatedconstant.REPLACEMENT_CHARACTER

crates/quickjs-wasm-rs/src/js_binding/value.rs
```diff
@@ -147,13 +149,60 @@ impl Value {
...
+                    res.push(REPLACEMENT_CHAR);
+
+                    // see https://simonsapin.github.io/wtf-8/#surrogate-byte-sequence
+                    let lone_surrogate =
+                        matches!(after_valid, [0xED, 0xA0..=0xBF, 0x80..=0xBF, ..]);
+                    // Rust's `Utf8Error` reports a lone surrogate as having an error_len of 1
+                    // which results in 3 replacement chars being inserted instead of 1. 1
```
saulecabrera, line 178: Do you know why this happens? By reading the documentation of `Utf8Error` this is not how I interpret this piece. 

https://doc.rust-lang.org/stable/std/str/struct.Utf8Error.html#method.error_len

> Decoding should resume after that sequence (after inserting a [U+FFFD REPLACEMENT CHARACTER](https://doc.rust-lang.org/stable/std/char/constant.REPLACEMENT_CHARACTER.html)) in case of lossy decoding.

PR #186, opened 2023-01-10
Add a build output to `javy` npm library

surma:
(I ran `prettier` for the first time. I recommend looking at the diff with `?w=1`)

So far we only provided the TS files. But if we want folks to be able to use their own build systems, we don't want to force them to have to deal with TypeScript. So I added `vite` to the npm library that compiles all our TS to JS with sidecar `.d.ts` files.

### DIFF
--- a/npm/javy/package.json
+++ b/npm/javy/package.json
@@ -2,13 +2,24 @@
 	"name": "javy",
 	"version": "0.0.1",
 	"description": "",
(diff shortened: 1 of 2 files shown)


saulecabrera (approved):
> I ran prettier for the first time.

I think we should include a `prettierrc`?

PR #188, opened 2023-01-11
More fixes for the WPT runner

surma:
As reported by @jeffcharles, there were more bugs in the WPT runner. This PR fixes them.

1.) I originally rewrote the WPT-invented script imports to ESM imports. However, ES module semantics gives each module its own scope. That means a variable declared at top-level of a module is effectively inaccessible, meaning certain variables got stripped. WPT on the other hand just inlines the referenced file. The rollup bundler script now should do the same as WPT.
2.) Failure of `setup()` methods are reported through a different callback than a test result (why? Who knows!). I added that second callback and report the failure.

### DIFF
--- a/wpt/reporter.js
+++ b/wpt/reporter.js
@@ -1,5 +1,12 @@
 export let failedTestCount = 0;
 
+function logFailure({ name, message, stack }) {
+  console.log("[FAIL]", name);
+  console.log(message);
+  console.log(stack);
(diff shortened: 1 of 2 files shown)


saulecabrera (approved):
Code LGTM.

jeffcharles (approved):
Thank you so much for doing this!

PR #196, opened 2023-01-16
Add workflow test for dylib that generates module

jeffcharles:
Adds another integration test that exercises the dylib. This time we compile the source code to bytecode in an isolated instance of the `javy_quickjs_provider.wasm` dylib. Then we generate a Wasm module using `wasm_encoder` containing a series of instructions to execute QuickJS bytecode that's been stored in a data section. This is more reflective of how the CLI and runtime environments will use the dylib.

I tried a few different approaches for how to structure the module generation to avoid relying on hard-coded indices as much as possible. I opted for the one here since it just introduces a single, not complicated abstraction to achieve that outcome. I could probably do something fancier with macros but this seems like a good enough start for now.

### DIFF
--- a/crates/cli/tests/module_generator/mod.rs
+++ b/crates/cli/tests/module_generator/mod.rs
@@ -1,0 +1,240 @@
+use anyhow::Result;
+use wasm_encoder::{
+    CodeSection, CustomSection, DataCountSection, DataSection, EntityType, ExportKind,
+    ExportSection, Function, FunctionSection, ImportSection, Instruction, MemorySection,
+    MemoryType, Module, TypeSection, ValType,
+};
+
+pub fn generate_module(bytecode: Vec<u8>, js_src: &str) -> Result<Vec<u8>> {
+    let mut module = Module::new();
+    let mut indices = Indices::new();
+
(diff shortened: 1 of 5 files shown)


crates/cli/tests/module_generator/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    exports.export("memory", ExportKind::Memory, indices.memory.unwrap());
+    exports.export("_start", ExportKind::Func, indices.start_fn.unwrap());
+    module.section(&exports);
+}
+
+fn add_data_count(module: &mut Module, count: u32) {
+    module.section(&DataCountSection { count });
+}
+
+fn add_code(module: &mut Module, indices: &Indices, bytecode_len: i32) {
```
saulecabrera, line 183: Can we add an internal comment that includes that wat that this function is generating? It will make it more manageable for newcomers.

saulecabrera: Actually it might be better to add this comment as part of `generate_module` function at the top.

saulecabrera (approved):

PR #198, opened 2023-01-17
Update GH actions actions to use v3 instead of v1

jeffcharles:
This should hopefully get rid of a bunch of deprecation warnings.

Currently seeing the following on our CI runs:
Screenshot 2023-01-17 at 13 40 24

[actions/upload-release-asset](https://github.com/actions/upload-release-asset) also appears to be unmaintained so we'll have to replace that with something else.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     name: Tests + Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
         with:
           submodules: true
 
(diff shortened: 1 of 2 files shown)


jeffcharles:
@saulecabrera is there a way I can test the release action without releasing a new version?

saulecabrera:
> @saulecabrera is there a way I can test the publish action without releasing a new version?

Not easily afaik. I've used https://github.com/nektos/act in the past. Perhaps you could try that?

jeffcharles:
Looks like _act_ requires Docker. I tried with Podman and am getting some weird errors. There are some workarounds in https://github.com/nektos/act/issues/303 but they seem to only work on Linux.

I'll just try to get this working in the repo so might be several more PRs as I iterate.

PR #199, opened 2023-01-18
Add dynamic linking support to CLI

jeffcharles:
Adds dynamic linking support to the Javy CLI with use of the `-d` flag. That is, the Wasm file output by Javy will be a Wasm module that contains QuickJS bytecode and a series of instructions to call into `javy_quickjs_provider.wasm` to execute that QuickJS bytecode. I'll update the README at a later date with instructions given a dynamically linked module can't really be used by anyone right now.

I've removed the dylib workflow test given the new integration test I've added essentially tests the same thing but uses the CLI to generate the Wasm module instead of doing it entirely in the test.

For the CLI build file, I chose not to add a separate environment variable to override the provider module location since I'm not clear on what purpose that would serve. I'm happy to add it if we see a reason to.

### DIFF
--- a/crates/cli/src/options.rs
+++ b/crates/cli/src/options.rs
@@ -12,6 +12,6 @@ pub struct Options {
 
     #[structopt(short = "d")]
     /// Creates a smaller module that requires a dynamically linked QuickJS provider Wasm
-    /// component to execute the JavaScript
+    /// module to execute the JavaScript
     pub dynamic: bool,
 }


crates/cli/src/module_generator.rs
```diff
@@ -36,7 +38,7 @@ use wasm_encoder::{
 //     (export "_start" (func 2))
 //     (data (;0;) "\02\03\0econsole\06log\18function.mjs\0e\00\06\00\a0\01\00\01\00\03\00\00\11\01\a2\01\00\00\008\de\00\00\00B\df\00\00\00\bd*$\01\00\cd(\c0\03\01\00")
 // )
-pub fn generate_module(bytecode: Vec<u8>, js_src: &str) -> Result<Vec<u8>> {
+pub fn generate_module(bytecode: Vec<u8>, js_src: &[u8]) -> Result<Vec<u8>> {
```
surma, line 41: [Super nit] One thing I didn’t spot earlier: Do we ever need the second memory?

surma (approved):
This is super exciting to see!!!

```
$ echo "console.log('hello')" > test.js                                                                                                                          
$ ./target/release/javy -d -o test.wasm test.js
$ ls -alh test.wasm
-rw-r--r--  1 surma  staff   348B Jan 19 12:29 test.wasm
$ wasmtime run --wasm-features multi-memory --preload javy_quickjs_provider_v1=target/wasm32-wasi/release/javy_quickjs_provider.wasm test.wasm
hello
```

crates/cli/src/custom_section.rs
```diff
@@ -0,0 +1,16 @@
...
+    brotli::enc::BrotliCompress(
+        &mut std::io::Cursor::new(source_code),
+        &mut compressed_source_code,
+        &brotli::enc::BrotliEncoderParams {
+            quality: 11,
+            ..Default::default()
+        },
+    )?;
+    Ok(compressed_source_code)
+}
```
saulecabrera, line 16: Can we change the name of this module to something else? Personally I feel that `custom_section` is not the most appropriate name, since it doesn't do anything custom section specific. Perhaps `compression` or something around those lines?

The only thing that is vaguely related to a custom section is `SOURCE_CODE_SECTION_NAME`, but that constant can live in `main.rs` which the module orchestrating the module creation.

crates/cli/src/options.rs
```diff
@@ -9,4 +9,7 @@ pub struct Options {
 
     #[structopt(short = "o", parse(from_os_str), default_value = "index.wasm")]
     pub output: PathBuf,
+
+    #[structopt(short = "d")]
+    pub dynamic: bool,
```
saulecabrera, line 14: Can we [add a description ](https://docs.rs/structopt/0.2.6/structopt/#help-messages)here so that the message is available when running with `--help`?

crates/cli/src/custom_section.rs
jeffcharles: I opted to go with `source_code_section`

crates/cli/src/module_generator.rs
surma: I mostly noticed it because neither `wasm2wat` nor `wasmtime` would swallow this without enabling multi memory.

PR #200, opened 2023-01-18
Remove install Rust steps from Github Actions

jeffcharles:
Looks like the `actions-rs/toolchain` Github action [is no longer maintained](https://www.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/) and is raising deprecation warnings. We _should_ be able to just use `rust-toolchain.toml` to configure our toolchain in Github Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,14 +15,6 @@ jobs:
         with:
           submodules: true
 
-      - name: Install Rust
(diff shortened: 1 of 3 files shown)


saulecabrera:
For reference, rustup is installed by default https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#rust-tools

PR #201, opened 2023-01-18
Attach gzipped dylib as part of publishing

jeffcharles:
Adds attaching `javy-quickjs_provider-${{ javy_version }}.wasm.gz` and its sha to releases.

This seems like a reasonable way to start publishing a dylib to use though it does tie Javy and `javy_quickjs_provider` versions together.

The name format is somewhat arbitrary and I'm okay with changing it if it makes sense. But I figured the `.wasm` extension should be enough of a hint about what architecture it's targeting to not need to include `wasm32-wasi` in the name.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -32,6 +32,44 @@ jobs:
           name: engine
           path: target/wasm32-wasi/release/javy_core.wasm
 
+      - name: Archive quickjs_provider
+        run: gzip -k -f target/release/javy_quickjs_provider.wasm && mv target/release/javy_quickjs_provider.wasm.gz javy-quickjs_provider-${{ github.event.release.tag_name }}.wasm.gz
+
+      - name: Upload quickjs_provider to artifacts
+        uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -32,6 +32,44 @@ jobs:
           name: engine
           path: target/wasm32-wasi/release/javy_core.wasm
 
+      - name: Archive quickjs_provider
+        run: gzip -k -f target/release/javy_quickjs_provider.wasm && mv target/release/javy_quickjs_provider.wasm.gz javy-quickjs_provider-${{ github.event.release.tag_name }}.wasm.gz
+
+      - name: Upload quickjs_provider to artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: javy-quickjs_provider-${{ github.event.release.tag_name }}.wasm.gz
```
saulecabrera, line 41: Should this be `javy_quickjs_provider-{version}`? To be consistent with the generated wasm name. Or even `javy_quickjs_provider_{version}` since I don't see a reason to have a `-` there.

saulecabrera (approved):

jeffcharles:
I'm going to hold off on this until we get the Wizer story for the provider figured out

PR #202, opened 2023-01-19
Wizer bytecode ahead of time

jeffcharles:


saulecabrera (approved):
Thanks for fixing this.

PR #205, opened 2023-01-20
Add new CLI flag to emit provider binary

surma:


crates/cli/src/main.rs
```diff
@@ -7,15 +7,25 @@ mod source_code_section;
...
 
+const QUICKJS_PROVIDER_MODULE: &[u8] = include_bytes!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/../../target/wasm32-wasi/release/javy_quickjs_provider.wasm"
+));
+
 fn main() -> Result<()> {
     let opts = Options::from_args();
 
+    if let Some(path) = &opts.provider {
```
surma, line 25: This is a bit iffy because it will only emit the binary if the rest of the invocation is valid. SO just `javy --emit-provider` fails. But I am not sure how to teach StructOpt to behave differently...

jeffcharles: Arguably some way of expressing a subcommand (for example, `javy emit-provider`) would be a way to do this so the flags are tied to a subcommand instead of Javy in general. But that's a more disruptive change.

crates/cli/src/options.rs
```diff
@@ -14,4 +14,9 @@ pub struct Options {
     /// Creates a smaller module that requires a dynamically linked QuickJS provider Wasm
     /// module to execute the JavaScript
     pub dynamic: bool,
+
+    #[structopt(long = "emit-provider")]
+    /// Emits an additional WebAssembly module that is needed to run
+    /// dynamically linked modules created with `-d`.
```
jeffcharles, line 20: Can we add some language to indicate we're expecting a path with this flag?

crates/cli/src/main.rs
saulecabrera: I'd prefer if we add a subcommand for this, since it is unrelated to the main command which is compilation.

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "stable"
+channel = "1.66.0"
```
surma, line 2: CI was failing because of the `wasm-opt` issue. This feels reasonable for now.

saulecabrera: Can we add a FIXME comment here and create an issue to track this? I'd like to get us back to stable as soon as we have a fix for this.

surma:
🤦  Good catches! Thanks both!

rust-toolchain.toml
saulecabrera: Sorry for the back and forth here, but with [this patch](https://github.com/Shopify/javy/pull/217), this is no longer required.

saulecabrera: I know we use cargo-wasi, but we never trigger release builds as far as I know, so `wasm-opt` should never get invoked.

PR #206, opened 2023-01-20
Add optional types to JS exceptions

jeffcharles:
Adds a public `JSError` error type to `quickjs-wasm-rs`. Each variant represents a different JS error type that QuickJS provides a function for throwing.  `wrap_callback` continues to expect an `anyhow::Error` to be returned from the callback but is augmented to attempt a downcast on any error returned and, if it's a `JSError`, throws the error for the variant used. Adding this enables us to throw `TypeError`s from `TextDecoder.decode` when decoding fails due to an invalid character when the `fatal` option is set to `true`, which enables more WPT cases to pass.

I opted for downcasting to a distinct error type since relying on the `msg` contents or `context` contents seemed to me like it would be an incorrect use of either field.

### DIFF
--- a/crates/quickjs-wasm-rs/Cargo.toml
+++ b/crates/quickjs-wasm-rs/Cargo.toml
@@ -17,7 +17,6 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 serde-transcode = { version = "1.1", optional = true }
 once_cell = "1.16"
-thiserror = "1.0"
 
 [dev-dependencies]
 quickcheck = "1"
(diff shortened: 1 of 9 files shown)


crates/quickjs-wasm-rs/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 serde-transcode = { version = "1.1", optional = true }
 once_cell = "1.16"
+thiserror = "1.0"
```
saulecabrera, line 20: I'd like to avoid this dependency if possible. I think it's still valuable to be conscious about the generated code size which we still want to keep as small as possible in all cases (dynamic vs static modules).

Since `anyhow` works with any type that implements `std::error::Error` is there any reason, on why we can't instead manually implement `Error` for `JSError`?

jeffcharles: That's fair. I originally wrote an implementation that manually implemented `Error` but it was messy. I added `thiserror` which cleaned it up, then cleaned up some other aspects of how it was implemented, and didn't revisit removing the dependency with the cleaned up code. The version without the dependency I just pushed up looks nicer than with `thiserror` so that's good.

surma (approved):

saulecabrera (approved):

PR #225, opened 2023-02-02
Handle null byte in error message better

jeffcharles:
Removes the last reference in Javy Core to `ext_js_exception`. I opted to truncate the error message if a null byte was detected and then append a string indicating we'd done that.


saulecabrera (approved):
Can we also remove `ext_js_exception` from `value.c`?

jeffcharles:
Yep. That will be a breaking change though but I agree we should do it. I'll do that as a follow-up PR.

PR #226, opened 2023-02-02
Add check publish workflow to PRs

jeffcharles:
Hopefully this will reduce the frequency of the publish workflow breaking.


saulecabrera (commented):
I wonder if we can [leverage re-using workflows here.](https://docs.github.com/en/actions/using-workflows/reusing-workflows) It might involve a bit of refactoring, but I think that it's worth the cost: if these go out of sync (the real workflow vs the test one) for whatever reason, we'd end up in a situation in which this flow would be invalid for our purposes.

jeffcharles:
This approach uses conditionals around the _upload to release_ logic rather than a reusable workflow. It's fewer files and workflows to keep track of which is nice.

PR #236, opened 2023-02-10
Add documentation for dynamic linking to README

jeffcharles:
I opted to repeat myself a bit around dynamically linked modules having special requirements and not working otherwise since I feel like this is something that a developer not paying close attention to could miss and then be very confused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,26 @@ In a runtime like Wasmtime, [wasmtime-wasi](
 https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/struct.WasiCtx.html#method.set_stdin)
 can be used to set the input and retrieve the output.
 
+## Creating and using dynamically linked modules
+
+In some scenarios, you may want or need to generate much smaller Wasm modules with Javy. Using the `-d` flag when invoking Javy will create a dynamically linked module which will have a much smaller file size than a statically linked module. Dynamically linked modules have special requirements that statically linked modules do not and will not execute in WebAssembly runtimes that do not meet these requirements.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,6 +118,26 @@ In a runtime like Wasmtime, [wasmtime-wasi](
 https://docs.rs/wasmtime-wasi/latest/wasmtime_wasi/struct.WasiCtx.html#method.set_stdin)
 can be used to set the input and retrieve the output.
 
+## Creating and using dynamically linked modules
+
+In some scenarios, you may want or need to generate much smaller Wasm modules with Javy. Using the `-d` flag when invoking Javy will create a dynamically linked module which will have a much smaller file size than a statically linked module. Dynamically linked modules have special requirements that statically linked modules do not and will not execute in WebAssembly runtimes that do not meet these requirements.
```
saulecabrera, line 123: I think we should probably clarify what a statically linked module means: embedding the entire JS engine.

README.md
```diff
@@ -118,6 +118,26 @@ In a runtime like Wasmtime, [wasmtime-wasi](
...
+
+To successfully instantiate and run a dynamically linked Javy module, the execution environment must provide a `javy_quickjs_provider_v1` namespace for importing that links to the exports provided by the `javy_quickjs_provider.wasm` module. Dynamically linked modules **cannot** be instantiated in environments that do not provide this import.
```
saulecabrera, line 125: I think we should also mention that dynamic modules are tied to QuickJS since they depend on its bytecode representation.

saulecabrera (approved):

PR #238, opened 2023-02-14
Enable promises behind a feature

jeffcharles:
The feature is off by default. It can be enabled by running `make cli CORE_FEATURES=experimental_event_loop` and tests can be run by invoking `make test-cli CORE_FEATURES=experimental_event_loop CLI_FEATURES=experimental_event_loop`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ cli: core
 
 core:
 		cd crates/core \
-				&& cargo build --release --target=wasm32-wasi \
+				&& cargo build --release --target=wasm32-wasi --features=$(CORE_FEATURES) \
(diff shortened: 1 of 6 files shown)


surma:
[Simplified the test runner](https://github.com/Shopify/javy/compare/surma/update-wpt?expand=1) since we can now rely on proises :D

saulecabrera:
I left an offline note to @jeffcharles, but I'll add here too for visibility: before landing this I'd like for us to ensure that unhandled rejected promises are behaving correctly, a preliminary test on my end shows that the current behaviour doesn't throw, which  is wrong afaik. So we might have to fix that.

surma:
This is probably also a reminder that we should run [test262](https://github.com/tc39/test262)

jeffcharles:
Discussed offline, we're going to put this behind a default off feature for now

PR #248, opened 2023-02-23
Refactor to use anyhow in benchmarks

jeffcharles:
I think this makes the code a little less noisy while I'm in here making other changes.

### DIFF
--- a/crates/cli/benches/benchmark.rs
+++ b/crates/cli/benches/benchmark.rs
@@ -1,6 +1,7 @@
+use anyhow::{anyhow, bail, Result};
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use num_format::{Locale, ToFormattedString};
-use std::{error::Error, fmt::Display, fs, path::Path, process::Command};
+use std::{fmt::Display, fs, path::Path, process::Command};
 use wasi_common::{
     pipe::{ReadPipe, WritePipe},
     WasiCtx,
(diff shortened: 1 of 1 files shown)


crates/cli/benches/benchmark.rs
```diff
@@ -22,16 +23,16 @@ impl Display for Function {
 }
 
 impl Function {
-    pub fn new(function_dir: &Path, js_path: &Path) -> Result<Function, Box<dyn Error>> {
+    pub fn new(function_dir: &Path, js_path: &Path) -> Result<Function> {
         let name = function_dir
             .file_name()
-            .ok_or("Path terminates in ..")?
+            .ok_or(anyhow!("Path terminates in .."))?
```
jianghong, line 29: Trying to understand why you need the `anyhow!` here.

Looking at the docs [here](https://docs.rs/anyhow/latest/anyhow/macro.anyhow.html), this `anyhow!` macro returns an `anyhow::Error` type.

Does the `anyhow!` cause this to return early or something?

jeffcharles: I get
```
the trait `std::error::Error` is not implemented for `str`
```
back from the compiler without performing a conversion.

That's a good question about why using a `&str` in `.ok_or()?` works with `Box<dyn std::error::Error>`.

jeffcharles: I also realized I didn't answer your question. It does not return early. `bail!` does that. It just takes a string or non-Anyhow error and converts it to an Anyhow error.

saulecabrera: > That's a good question about why using a &str in .ok_or()? works with Box<dyn std::error::Error>.

Because &str implements it That's a good question about why using a &str in .ok_or()? works with Box<dyn std::error::Error>

saulecabrera: > Trying to understand why you need the anyhow! here.

Mainly because by using anyhow we are using its `Result` type, which expects `anyhow::Error` in the Error arm of the `Result` enum and `anyhow!` is a convenient way of creating an error from a `&str`.

jeffcharles: > That's a good question about why using a &str in .ok_or()? works with Box<dyn std::error::Error>.

Looks like `Box<dyn Error + 'static>` and `Box<dyn Error + Send + Sync + 'a>` have a `From<&str>` implementation. `anyhow::Error` does not have a similar `From<&str>` implementation.

https://doc.rust-lang.org/std/error/trait.Error.html#impl-From%3C%26str%3E-for-Box%3Cdyn%20Error%20+%20%27static%2C%20Global%3E for details

jianghong: Thanks!

PR #249, opened 2023-02-24
Refactor benchmark funcs into function cases

jeffcharles:
In preparation of adding benchmarks for dynamically linked cases, I wanted to simplify how we benchmark different ways of compiling and executing the same function. The idea is instead of invoking a different run method on each `Function` for every combination of ways of compiling the Function, we instead use a `FunctionCase` that's configured for a particular way of compiling a Function that performs the expected behaviour when its `setup` and `run` methods are invoked. Without this, we'd have to add new run functions and possibly new setup functions for every combination of parameters (for example, `run_uncompiled_dynamic`, `run_precompiled_dynamic`, `run_uncompiled_static`, `run_precompiled_static`) which starts to get messy. This does move some complexity into `FunctionCase`'s internals but we can use appropriate data structures and potentially delegation inside the struct to manage that complexity while keeping the interface into the struct relatively small.

### DIFF
--- a/crates/cli/benches/benchmark.rs
+++ b/crates/cli/benches/benchmark.rs
@@ -117,6 +117,14 @@ pub fn criterion_benchmark(c: &mut Criterion) {
             )
             .unwrap(),
         );
+        function_cases.push(
+            FunctionCase::new(
+                Path::new("benches/functions/complex_discount"),
+                Path::new("dist/function.js"),
+                precompiled,
+            )
+            .unwrap(),
+        );
     }
 
     for function_case in function_cases {
--- a/crates/cli/benches/benchmark.rs
+++ b/crates/cli/benches/benchmark.rs
@@ -23,9 +23,9 @@ impl Display for FunctionCase {
             f,
             "{} {}",
             if self.precompiled_elf_bytes.is_some() {
-                "precompiled"
+                "ahead of time"
             } else {
-                "uncompiled"
+                "just in time"
             },
             self.name
         )
@@ -47,10 +47,9 @@ impl FunctionCase {
         let engine = Engine::default();
         let wasm_bytes = fs::read(wasm_path)?;
 
-        let precompiled_elf_bytes = if let Compilation::Precompiled = compilation {
-            Some(Module::new(&engine, &wasm_bytes)?.serialize()?)
-        } else {
-            None
+        let precompiled_elf_bytes = match compilation {
+            Compilation::AheadOfTime => Some(Module::new(&engine, &wasm_bytes)?.serialize()?),
+            Compilation::JustInTime => None,
         };
         let module_size = precompiled_elf_bytes
             .as_ref()
@@ -108,7 +107,7 @@ impl FunctionCase {
 
 pub fn criterion_benchmark(c: &mut Criterion) {
     let mut function_cases = vec![];
-    for compilation in [Compilation::Uncompiled, Compilation::Precompiled] {
+    for compilation in [Compilation::JustInTime, Compilation::AheadOfTime] {
         function_cases.push(
             FunctionCase::new(
                 Path::new("benches/functions/simple_discount"),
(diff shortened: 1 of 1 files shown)


crates/cli/benches/benchmark.rs
```diff
@@ -101,48 +107,26 @@ impl Function {
...
-        );
-
-        let serialized_module = function.compile().unwrap();
-        println!(
-            "Size of precompiled module for {}: {} bytes",
-            function,
-            serialized_module.len().to_formatted_string(&Locale::en)
+    let mut function_cases = vec![];
+    for precompiled in [false, true] {
+        function_cases.push(
```
jianghong, line 112: I think you missed pushing the complex_discount case here

jianghong (commented):
makes sense

crates/cli/benches/benchmark.rs
jeffcharles: Yes, yes I did 🤦

jianghong (approved):

crates/cli/benches/benchmark.rs
```diff
@@ -34,46 +44,42 @@ impl Function {
...
 
-        Ok(Function {
+        let engine = Engine::default();
+        let wasm_bytes = fs::read(wasm_path)?;
+
+        let precompiled_elf_bytes = if precompiled {
+            Some(Module::new(&engine, &wasm_bytes)?.serialize()?)
+        } else {
+            None
+        };
```
saulecabrera, line 54: Can we add different constructors here (`Function::precompiled` and `Func::new`) instead of passing  a boolean flag? Without enough context on the underlying implementation it might be hard to tell what the flag is for.

saulecabrera (approved):

crates/cli/benches/benchmark.rs
jeffcharles: The reason I structured it like this is to avoid a combinatorial explosion every time we add a new dimension we want to test on.

E.g., if we were to add a third dimension called `forked` or `unforked` (just an example because I can't think of what the next dimension will be) I don't want

```rust
FunctionCase::new_static_uncompiled_nonforking(...)?;
FunctionCase::new_static_uncompiled_nonforking(...)?;
FunctionCase::new_dynamic_uncompiled_nonforking(...)?;
FunctionCase::new_dynamic_uncompiled_nonforking(...)?;
FunctionCase::new_static_precompiled_nonforking(...)?;
FunctionCase::new_static_precompiled_nonforking(...)?;
FunctionCase::new_dynamic_precompiled_nonforking(...)?;
FunctionCase::new_dynamic_precompiled_nonforking(...)?;
FunctionCase::new_static_uncompiled_forking(...)?;
FunctionCase::new_static_uncompiled_forking(...)?;
FunctionCase::new_dynamic_uncompiled_forking(...)?;
FunctionCase::new_dynamic_uncompiled_forking(...)?;
FunctionCase::new_static_precompiled_forking(...)?;
FunctionCase::new_static_precompiled_forking(...)?;
FunctionCase::new_dynamic_precompiled_forking(...)?;
FunctionCase::new_dynamic_precompiled_forking(...)?;
```

because it's super easy to miss a case and not even realize it.

What I'm open to is switching it to an enum instead of a bool since that's a totally fair criticism that it's difficult to determine what the boolean represents!

So we could do something like:

```rust
for linking in [Linking::Static, Linking::Dynamic] {
  for compilation in [Compilation::Precompiled, Compilation::Uncompiled] {
    for forking in [Forking::Forking, Forking::NonForking] {
      FunctionCase::new(..., linking, compilation, forking)?;
      FunctionCase::new(..., linking, compilation, forking)?;
    }
  }
}
```

saulecabrera (approved):
LGTM; left a small nit and a question. Feel free to merge once those are addressed.

crates/cli/benches/benchmark.rs
saulecabrera: Oh makes sense, I misread that!

PR #250, opened 2023-02-27
Enable dependabot

jeffcharles:
Enabling dependabot so we're better able to keep up with new versions of dependencies.

I don't see a way to test the configuration unfortunately. I also confirmed that the Github Actions directory needs to be set to `/` according to https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory.

### DIFF
--- a/.github/.dependabot
+++ b/.github/.dependabot
@@ -3,19 +3,19 @@ updates:
   - package-ecosystem: "cargo"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
     open-pull-requests-limit: 100
   - package-ecosystem: npm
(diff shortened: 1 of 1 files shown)


.github/.dependabot
```diff
@@ -0,0 +1,21 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
saulecabrera, line 6: I feel that this might be too aggressive and cause a lot of noise -- what do you think of weekly / monthly?

jianghong (approved):

PR #253, opened 2023-02-27
Use --all-targets flag for Clippy

jeffcharles:
This should help us keep Javy test and benchmarking code more clean and readable by adding some additional static analysis.

### DIFF
--- a/crates/quickjs-wasm-rs/src/serialize/de.rs
+++ b/crates/quickjs-wasm-rs/src/serialize/de.rs
@@ -245,16 +245,14 @@ mod tests {
     fn test_null() {
         let context = Context::default();
         let val = context.null_value().unwrap();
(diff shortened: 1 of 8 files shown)


crates/quickjs-wasm-rs/src/serialize/de.rs
```diff
@@ -245,16 +245,14 @@ mod tests {
     fn test_null() {
         let context = Context::default();
         let val = context.null_value().unwrap();
-        let actual = deserialize_value::<()>(val);
-        assert_eq!((), actual);
+        deserialize_value::<()>(val);
```
jeffcharles, line 248: This will always return `()` because of the generic type being passed so Clippy complains about the let binding and the assert check. I think as long as this runs without panicking, it should be fine IMO.

PR #254, opened 2023-02-28
Use package flag in Makefile

jeffcharles:
This change has us use the `--package` flag instead of `cd ... && ... && cd -` pattern. I find this reduces noise in the Makefile. I also updated the `npm` commands to use the `--prefix` flag to avoid having to change directories. I also updated the `docs` target to generate and open docs for both the CLI and Javy Core instead of just Javy Core.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,82 +8,61 @@ install:
 	cargo install --path crates/cli
 
 bench: cli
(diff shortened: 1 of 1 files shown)


jianghong (approved):
Nice!

PR #262, opened 2023-03-06
Add a github issue template for bug reports and ideas

jianghong:
Thoughts on if the `idea.yml` template is needed?


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,18 @@
+---
+name: Bug report
+about: File a bug report
+title: 
+labels: 
+assignees: 
+
+---
+
+Operating system: ...
```
jeffcharles, line 10: Might also be worth asking for their processor architecture

jeffcharles (commented):
We do get people who sometimes ask questions so might be worth having a template for them. Especially since we can automatically apply a label.

jeffcharles (approved):

PR #268, opened 2023-03-07
Compile bytecode as module instead of as global

jeffcharles:
We want to offer the option to compile bytecode as a JS module instead of using the JavaScript global environment and to switch Javy to use a JS module.

I'm keeping the existing `compile_global` function in `quickjs-wasm-rs` and having it compile to use the global environment and adding a `compile_module` function to compile to a JS module.

I've tweaked the tests a bit too. `await` doesn't behave that differently between global and module but importing a module does. So I've tweaked the tests to use that as an example of a syntax error in global mode and a different error in module mode. I've also added a test to check that module compilation doesn't leak variables into the global object.


jianghong (approved):
Makes sense.

saulecabrera (approved):

PR #277, opened 2023-03-10
Error if there are pending items on the event queue

jeffcharles:
We've received feedback that the current behaviour of successfully exiting when items there are items enqueued on to the event loop that will never be executed is unexpected and they would prefer if the execution trapped with a helpful error message instead. This PR:
- Adds an `is_pending` to `Context` to return whether there are pending jobs on the event queue
- Changes the run bytecode logic to panic if there are pending jobs on the event queue with an exception if we're running the Web Platform Tests suite as the suite enqueues a promise as part of its setup
- Disables a WPT suite that enqueues events

The check if we're running the WPT suite is done through a WASI environment variable. I opted to use this approach because:
- I couldn't figure out a reasonable way to change how we run the WPT suite so that it doesn't enqueue the promise
- we won't have to use a special build of Javy for the WPT suite
- we don't have to touch the Javy CLI including adding support for additional arguments or env vars and plumbing it through to the core logic

I did take a look at the benchmarks to see if the WASI call to retrieve the env var value would have a performance impact and the results indicated no change in performance between main and this branch.

### DIFF
--- a/crates/core/src/execution.rs
+++ b/crates/core/src/execution.rs
@@ -1,10 +1,19 @@
-use anyhow::Result;
+use std::env;
+
+use anyhow::{bail, Result};
 use quickjs_wasm_rs::Context;
 
 pub fn run_bytecode(context: &Context, bytecode: &[u8]) -> Result<()> {
     context.eval_binary(bytecode)?;
     if cfg!(feature = "experimental_event_loop") {
         context.execute_pending()?;
+    } else if context.is_pending() && !is_running_wpt_suite() {
+        // WPT enqueues a promise as part of its suite setup and we can't change that behaviour so don't error
+        bail!("Adding tasks to the event queue is not supported");
     }
     Ok(())
+}
+
+fn is_running_wpt_suite() -> bool {
+    env::var("JAVY_WPT").is_ok()
 }
--- a/wpt/package.json
+++ b/wpt/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "bundle": "rollup -c rollup.config.js runner.js",
     "javy": "../target/release/javy compile -o bundle.wasm bundle.js",
-    "wasmtime": "wasmtime bundle.wasm",
+    "wasmtime": "wasmtime --env JAVY_WPT=1 bundle.wasm",
     "test": "npm run bundle && npm run javy && npm run wasmtime"
   },
   "devDependencies": {
--- a/wpt/test_spec.js
+++ b/wpt/test_spec.js
@@ -19,9 +19,9 @@ export default [
   // { // FIXME requires `encodeInto` support
   //   testFile: "upstream/encoding/encodeInto.any.js",
   // },
-  {
-    testFile: "upstream/encoding/replacement-encodings.any.js",
-  },
+  // { // FIXME requires the event loop
+  //   testFile: "upstream/encoding/replacement-encodings.any.js",
+  // },
   {
     testFile: "custom_tests/textdecoder-arguments.any.js",
   },
--- a/crates/cli/tests/integration_test.rs
+++ b/crates/cli/tests/integration_test.rs
@@ -80,6 +80,19 @@ fn test_promises() {
     assert_eq!("\"foo\"\"bar\"".as_bytes(), output);
 }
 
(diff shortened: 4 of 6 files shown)


wpt/test_spec.js
```diff
@@ -19,9 +19,9 @@ export default [
   // { // FIXME requires `encodeInto` support
   //   testFile: "upstream/encoding/encodeInto.any.js",
   // },
-  {
-    testFile: "upstream/encoding/replacement-encodings.any.js",
-  },
+  // { // FIXME requires the event loop
+  //   testFile: "upstream/encoding/replacement-encodings.any.js",
```
jianghong, line 23: It's unclear to me why this was needed to be commented out, could you clarify?

wpt/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "bundle": "rollup -c rollup.config.js runner.js",
     "javy": "../target/release/javy compile -o bundle.wasm bundle.js",
-    "wasmtime": "wasmtime bundle.wasm",
+    "wasmtime": "wasmtime --env JAVY_WPT=1 bundle.wasm",
```
jianghong, line 9: I don't know much about how the `wpt `stuff is built and ran yet but could we run it with the `experimental_event_loop` feature turned on? Would that help avoid needing to pass and checking this `JAVY_WPT`?

wpt/test_spec.js
jeffcharles: This actually doesn't need to be commented out any more since I disabled the error message with pending jobs. Technically the tests aren't completing properly but I can't figure out a way to have these tests error but have the suite setup to not cause a panic.

crates/core/src/execution.rs
```diff
@@ -1,10 +1,19 @@
...
+    } else if context.is_pending() && !is_running_wpt_suite() {
+        // WPT enqueues a promise as part of its suite setup and we can't change that behaviour so don't error
+        bail!("Adding tasks to the event queue is not supported");
     }
     Ok(())
 }
+
+fn is_running_wpt_suite() -> bool {
+    env::var("JAVY_WPT").is_ok()
+}
```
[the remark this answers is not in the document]
jianghong: What are we hoping to do with the `experimental_event_loop` feature? Is it experimental just for testing for now, but do we eventually want to promote it to the default behaviour? If so, I feel like running the WPT suite with it enabled should be fine since that's where we want to be eventually, but I could be missing something in this assumption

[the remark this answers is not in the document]
saulecabrera: > I want to run the WPT suite with Javy in state (2) because (1) potentially changes Javy's behaviour under test and (3) causes a panic on suite setup. Unfortunately it's a pain to figure out how to make that work without some sort of compromise somewhere.

Isn't 3 the "correct" behaviour? I'd argue that if WPT is enqueuing jobs, we should be providing the right behaviour at the core layer and not something that we consider "right". That's my main reason to suggest enabling the event loop. Also, can you provide a concrete example for:

> If we use a build of Javy with the experimental event loop flag enabled, what we're testing is something that potentially behaves differently in material ways. The tests could have different results than if it were disabled. I'd prefer not running the event loop because of that.
 
I'm failing to see how this could be an issue given that we are relying on the event loop for our own tests: we are not preventing new jobs from getting added, it's just that we are not flushing the queue.

-- 

As a bottom line:

(i) we are changing the behaviour here by saying: "if we are running WPT tests, we don't check for pending jobs", which is non-standard afaik, so this warrants a different build IMO: instead of using an env var, we might as well introduce a new feature, such as `wpt`. I'd be more comfortable with that. I do worry that the complexity of build types is going up, but I think it's less error prone and more explicit than an env var.

(ii) Given that our tests rely on the queue, I'd argue that getting the right level of compatibility for the event loop should be prioritized. Without a concrete example of how the event loop can introduce different results in our tests, I still think that enabling the event loop in the tests is better than introducing custom behaviour.

jeffcharles:
> For when we release: Would you consider this a major change?

We can bump the minor version since we're 0.x. JS code that was working could stop working if it created pending jobs.

saulecabrera (approved):
LGTM; thanks for all your patience iterating on this!

PR #279, opened 2023-03-13
Reduce the number of host calls when logging

jeffcharles:
While investigating something related to failures when calling `console.error`, I noticed our current implementation is making more host calls than necessary when writing logs. Each call to `write` on `stream` will invoke a host call since writes to the stderr stream are not buffered.

This change ensures we only make one one host call when invoking `console.error` which should improve runtime performance.

### DIFF
--- a/crates/core/src/globals.rs
+++ b/crates/core/src/globals.rs
@@ -70,15 +70,18 @@ where
     T: Write + 'static,
 {
     move |ctx: &Context, _this: &Value, args: &[Value]| {
+        // write full string to in-memory destination before writing to stream since each write call to the stream
+        // will invoke a hostcall
+        let mut log_line = String::new();
         for (i, arg) in args.iter().enumerate() {
             if i != 0 {
-                write!(stream, " ")?;
+                log_line.push(' ');
             }
 
-            stream.write_all(arg.as_str()?.as_bytes())?;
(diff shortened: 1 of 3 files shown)


crates/core/src/globals.rs
```diff
@@ -70,15 +70,18 @@ where
     T: Write + 'static,
 {
     move |ctx: &Context, _this: &Value, args: &[Value]| {
+        // write full string to in-memory destination before writing to stream since each write call to the stream
+        // will invoke a hostcall
```
saulecabrera, line 74:
```suggestion
    // Write full string to in-memory destination before writing to stream since each write call to the stream
    // will invoke a hostcall.
    ```

saulecabrera:
> which should improve runtime performance.

I think we should run some benchmarks as part of this change to have a concrete idea of the improvements that we're referring to here.

saulecabrera:
> which should improve runtime performance.

I think we should run some benchmarks as part of this change to have a concrete idea of the improvements that we're referring to here.

jianghong (approved):
Nice find

jeffcharles:
Added a benchmark, I'm seeing a 9% to 17% improvement for 100 `console.error` calls with multiple arguments

PR #293, opened 2023-04-03
Emphasize small binary size use case

jeffcharles:
We want the readme to emphasize how producing small binaries is an important use case.

Do these two changes do that sufficiently for our purposes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## About this repo
 
-**Introduction**: Run your JavaScript on WebAssembly. Javy takes your JavaScript code, and executes it in a WebAssembly embedded JavaScript runtime.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 ## About this repo
 
-**Introduction**: Run your JavaScript on WebAssembly. Javy takes your JavaScript code, and executes it in a WebAssembly embedded JavaScript runtime.
+**Introduction**: Run your JavaScript on WebAssembly. Javy takes your JavaScript code, and executes it in a WebAssembly embedded JavaScript runtime. Javy can create _very_ small Wasm modules in the 1 to 16 KB range with use of dynamic linking.
```
saulecabrera, line 7: What do you think of adding the baseline measurement for static modules here as well?

saulecabrera (approved):
Left a small comment, else LGTM

Issue #295, opened 2023-04-04
quickjs-wasm-rs improvements


jianghong:
## Rename `Value` and `Context`
As we plan to introduce an enum that represents a `JSValue` (see below), this rename will help us disambiguate which level of representation a type is. It also helps clarify the responsibility of the type by making it more explicit that it’s simple a Wrapper/Ref.

Renames:
* `Value` -> `JSValueRef`
* `Context` -> `JSContextRef`

## Enum representation for `JSValue`
Introduce an enum that represents the underlying `quickjs::JSValue`
```rust
pub enum JSValue {
    Undefined,
    Null,
    Bool(bool),
    Int(i32),
    Float(f64),
    String(String),
    Array(Vec<JSValue>),
    Object(HashMap<String, JSValue>),
    MutArrayBuffer(*mut u8, usize), // used for readSync. need to hold a raw pointer to quickjs memory to write directly to
}

impl JSContextRef {
  ...
  pub fn deserialize(&self, val: &JSValueRef) -> Result<JSValue> { ... }
  pub fn serialize(&self, val: JSValue) -> Result<JSValueRef> { ... }
  ...
}
```
Motivations:
1) The bindings become more ergonomic, they become truly Rust based, allowing consumers to deal only with high-level and safe Rust types. 

2) The deserialization and serialization are moved to a single place, rather than having independent serialization calls at each JS <> Rust touchpoint, which becomes error prone.

## `CallbackArg` 
Introduce a `CallbackArg` struct that will be used when defining callback functions. This is an optimization so that deserialization happens lazily, only if the `CallbackArg` is needed and used. 

```rust
pub struct CallbackArg<'a> {
    inner: JSValueRef<'a>,
}

impl CallbackArg<'_> {
    pub fn value(&self) -> Result<JSValue> {
        self.value.context.deserialize(&self.inner)
    }
}


// example callback function
fn foo() -> impl FnMut(&JSContextRef, &CallbackArg, &[CallbackArg]) -> anyhow::Result<JSValue> {
    move |_: &JSContextRef, _this: &CallbackArg, args: &[CallbackArg]| {
        // deserialize args to get native Rust types
        let arg0 = args[0].value().as_i32()?;
        let arg1 = args[1].value().as_string()?;
        let arg2 = args[3].value()?.as_bool()?;
        
       // do things with args 
        ... 

        // serialize final result happens when the result gets passed back
        Ok(JSValue::String(result))
    }
}

// set function on global
global.set_property(
[...]

PR #298, opened 2023-04-05
Rename Context -> JSContextRef, Value -> JSValueRef

jianghong:
Part of #295 

# Motivation

As we plan to introduce an new enum named `JSValue` (see #295), this rename will help us disambiguate things. It also helps clarify the responsibility of the type by making it more explicit that it’s simple a Wrapper/Ref.

So at the end of all the improvements, the available types would look something like this

- `enum JSValue`: A high level representation of `JSValue` with native rust types
- `JSValueRef`: Contains the raw pointer to a `quickjs_wasm_sys::JSValue`
- `JSContextRef`: Contains the raw pointer to a `quickjs_wasm_sys::JSContext`
- `quickjs_wasm_sys::JSValue`: The quickjs type
- `quickjs_wasm_sys::JSContext`: The quickjs type

# Regarding version
We decided to start using the suffix `alpha.x` to indicate a breaking change for when we pushing changes to the repo but have not published a new version to `crates.io` yet. So given this PR is a breaking change, I've set the new version to be `1.0.0-alpha.1` (assuming we are aligned that we are working towards a `1.0.0` release)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ docs:
 	cargo doc --package=javy-core --open --target=wasm32-wasi
 
 test-quickjs-wasm-rs:
-	cargo wasi test --package=quickjs-wasm-rs --features json -- --nocapture
+	cargo wasi test --package=quickjs-wasm-rs --features json,messagepack -- --nocapture
 
 test-core:
 	cargo wasi test --package=javy-core -- --nocapture
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1283,7 +1283,7 @@ dependencies = [
 
 [[package]]
 name = "quickjs-wasm-rs"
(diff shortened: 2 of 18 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ docs:
 	cargo doc --package=javy-core --open --target=wasm32-wasi
 
 test-quickjs-wasm-rs:
-	cargo wasi test --package=quickjs-wasm-rs --features json -- --nocapture
+	cargo wasi test --package=quickjs-wasm-rs --features json,messagepack -- --nocapture
```
jianghong, line 28: I noticed we weren't testing the messagepack module at all

surma: Are we still using messagepack?

PR #305, opened 2023-04-17
Update Wizer and Wasmtime

jeffcharles:
Update Wizer to a version which uses Wasmtime 3.x which should resolve a security alert involving Wasmtime.


crates/cli/tests/runner/mod.rs
```diff
@@ -20,7 +20,7 @@ pub struct Runner {
 pub struct RunnerError {
     pub stdout: Vec<u8>,
     pub stderr: Vec<u8>,
-    pub trap: Trap,
+    pub err: anyhow::Error,
```
jianghong, line 23: Just curious why you decided to change this

jianghong (approved):
LGTM, non blocking question

PR #306, opened 2023-04-18
Test for fuel consumption in integration tests

jianghong:
Given Shopify's relatively strict requirements around fuel usage, I think it's important that any changes we make do not increase fuel consumption significantly as changes here are out of the control of a Shopify partner so we need to be extra careful.

This PR adds some baseline values to assert that fuel consumed in our integration tests do not exceed a certain threshold. I'm currently defining that threshold as 10% but open to discussing other values. This threshold value needs to be large enough to account for fuel differences between different operating systems so the test does not end up being inconsistent

Example of a failed test
```
thread 'test_identity' panicked at 'fuel_consumed (37907) was not within 1.00% of the target_fuel value (39907)', crates/cli/tests/integration_test.rs:142:5
```

### DIFF
--- a/crates/cli/tests/integration_test.rs
+++ b/crates/cli/tests/integration_test.rs
@@ -7,48 +7,53 @@ use std::str;
 fn test_identity() {
     let mut runner = Runner::default();
 
-    let (output, _) = run_with_u8s(&mut runner, 42);
+    let (output, _, fuel_consumed) = run_with_u8s(&mut runner, 42);
     assert_eq!(42, output);
+    assert_fuel_consumed_within_threshold(37907, fuel_consumed);
 }
 
 #[test]
 fn test_fib() {
     let mut runner = Runner::new("fib.js");
 
-    let (output, _) = run_with_u8s(&mut runner, 5);
+    let (output, _, fuel_consumed) = run_with_u8s(&mut runner, 5);
     assert_eq!(8, output);
+    assert_fuel_consumed_within_threshold(59_822, fuel_consumed);
 }
 
 #[test]
 fn test_recursive_fib() {
     let mut runner = Runner::new("recursive-fib.js");
 
-    let (output, _) = run_with_u8s(&mut runner, 5);
+    let (output, _, fuel_consumed) = run_with_u8s(&mut runner, 5);
     assert_eq!(8, output);
+    assert_fuel_consumed_within_threshold(61_617, fuel_consumed);
 }
 
 #[test]
 fn test_str() {
     let mut runner = Runner::new("str.js");
 
-    let (output, _) = run(&mut runner, "hello".as_bytes());
+    let (output, _, fuel_consumed) = run(&mut runner, "hello".as_bytes());
     assert_eq!("world".as_bytes(), output);
+    assert_fuel_consumed_within_threshold(142_849, fuel_consumed);
 }
 
 #[test]
 fn test_encoding() {
     let mut runner = Runner::new("text-encoding.js");
 
-    let (output, _) = run(&mut runner, "hello".as_bytes());
+    let (output, _, fuel_consumed) = run(&mut runner, "hello".as_bytes());
     assert_eq!("el".as_bytes(), output);
+    assert_fuel_consumed_within_threshold(258_852, fuel_consumed);
 
-    let (output, _) = run(&mut runner, "invalid".as_bytes());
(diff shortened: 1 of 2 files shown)


jianghong:
Bumped to 10% threshold

jeffcharles:
I did a quick check locally and I'm also getting different fuel consumption counts:

```
thread 'test_identity' panicked at 'fuel_consumed (37903) was not within 0.00% of the target_fuel value (37907)', crates/cli/tests/integration_test.rs:142:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test test_identity ... FAILED
thread 'test_recursive_fib' panicked at 'fuel_consumed (62207) was not within 0.00% of the target_fuel value (61617)', crates/cli/tests/integration_test.rs:142:5
test test_recursive_fib ... FAILED
thread 'test_readme_script' panicked at 'fuel_consumed (284788) was not within 0.00% of the target_fuel value (284736)', crates/cli/tests/integration_test.rs:142:5
test test_readme_script ... FAILED
thread 'test_logging' panicked at 'fuel_consumed (21729) was not within 0.00% of the target_fuel value (22296)', crates/cli/tests/integration_test.rs:142:5
thread 'test_encoding' panicked at 'fuel_consumed (259025) was not within 0.00% of the target_fuel value (258852)', crates/cli/tests/integration_test.rs:142:5
thread 'test_fib' panicked at 'fuel_consumed (60228) was not within 0.00% of the target_fuel value (59822)', crates/cli/tests/integration_test.rs:142:5
test test_logging ... FAILED
test test_fib ... FAILED
test test_encoding ... FAILED
thread 'test_str' panicked at 'fuel_consumed (142656) was not within 0.00% of the target_fuel value (142849)', crates/cli/tests/integration_test.rs:142:5
test test_str ... FAILED
```

Now I'm thinking it may not be OS and architecture dependent but something else may be going on.

jeffcharles (approved):
I can't narrow down why the fuel counts are different given we _should_ have an identical environment for building Jayv and executing it.

jeffcharles:
A few things I investigated and confirmed:
- That we're using the same version of Wizer (confirmed we're both using 1.6.0)
- That we're using the same Rust version (1.68.2)
- That we're using the same wasi-sdk version (initially different but now both using version 19 but still seeing divergence between both of our machines and CI)
- That we're using an Arm64 MacOS build of Javy

Areas that will take more investigation but probably don't need to happen now:
- Wizering the core crate invokes a number of WASI functions (`random_get`, `environ_size_get`, `environ_get`, `clock_time_get`) and the results of those function invocations may result in different numbers of instructions being executed

saulecabrera:
Just wanted to drop some findings about this issue:

The `javy_core.wasm` that gets compiled differs between target architectures, for example:
  * In macOS it has 1,205,985 bytes
  * In Linux it has 1, 391, 287 bytes

In theory the difference above shouldn't matter for the released artifacts, since we only bundle the `macOS` version of `javy_core.wasm` in the CLI; but local testing vs in CI this is not the case. 

Another interesting thing is that the final module generated by javy has differences, which makes me think that at the CLI level, we are potentially baking some architecture specific code. The differences are mainly in the data section of the module:
Screenshot 2023-04-21 at 2 19 49 PM

jeffcharles:
> Another interesting thing is that the final module generated by javy has differences

Yeah the final wizened Javy QuickJS Wasm module is unfortunately going to be different each time the Wizening runs. In the Wizer initialization function, we create a QuickJS context. And [part of creating the QuickJS context](https://github.com/Shopify/javy/blob/main/crates/quickjs-wasm-sys/quickjs/quickjs.c#L51034) is setting a seed for `Math.random()` which gets the current time and performs some math operations on it to set the random seed. The random seed would be different from build to build and would be placed in the data section.

jianghong:
> javy_core module compiled in macos to report fuel?

Sorry I'm not sure what you mean by this could you expand?

jeffcharles:
> I think that the data sections should not have an impact on fuel

I don't think the data sections are having an impact on instruction count either since the instruction counts seem consistent on my machine from build to build. That said, if a data section did encode a hashmap that's been built based off of a seed obtained by `random_get`, the number of operations to fetch the value for a given key may change (the no collision case would have fewer operations compared to the case of there being a hash collision), but I don't think that's happening here.

> The local vs CI issue remains though

Yes. And also the issue that Jackson and I were seeing different instruction counts between the two of our machines despite having identical setups including the same CPU architecture.

PR #329, opened 2023-04-26
Update to Wasmtime 8

### LINKED ISSUE #328: Update to latest version of wasmtime
## What is the idea?
Update our wasmtime dependency to `8.0.0`. 

I did a [quick try](https://github.com/Shopify/javy/compare/jh/update-wasmtime-8?expand=1) and it seems like we need to do some resolution

```
cargo build --package=javy-core --release --target=wasm32-wasi --features=
    Updating crates.io index
error: failed to select a version for `wasi-common`.
    ... required by package `wasi-cap-std-sync v4.0.0`
    ... which satisfies dependency `wasi-cap-std-sync = "^4.0.0"` of package `wizer v1.6.1-beta.4 (https://github.com/bytecodealliance/wizer#b3b6ed48)`
    ... which satisfies git dependency `wizer` (locked to 1.6.1-beta.4) of package `javy-cli v0.6.0 (/Users/jackson/src/github.com/Shopify/javy/crates/cli)`
versions that meet the requirements `=4.0.0` are: 4.0.0

the package `wasi-common` links to the native library `wasi-common-19`, but it conflicts with a previous package which links to `wasi-common-19` as well:
package `wasi-common v8.0.0`
    ... which satisfies dependency `wasi-common = "^8.0"` of package `javy-cli v0.6.0 (/Users/jackson/src/github.com/Shopify/javy/crates/cli)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the links ='wasi-common' value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `wasi-common` which could resolve this conflict
make: *** [core] Error 101
```

jeffcharles:
Fixes #328. Updating Wizer so we get Wasmtime 8.

### DIFF
--- a/crates/cli/tests/runner/mod.rs
+++ b/crates/cli/tests/runner/mod.rs
@@ -43,12 +43,13 @@ struct StoreContext {
 
 impl StoreContext {
     fn new(input: &[u8], capacity: usize) -> Self {
-        let mut wasi = WasiCtxBuilder::new().inherit_stdio().build();
         let wasi_output = WritePipe::new_in_memory();
         let log_stream = WritePipe::new(LogWriter::new(capacity));
-        wasi.set_stdout(Box::new(wasi_output.clone()));
-        wasi.set_stdin(Box::new(ReadPipe::from(input)));
-        wasi.set_stderr(Box::new(log_stream.clone()));
+        let wasi = WasiCtxBuilder::new()
+            .stdout(Box::new(wasi_output.clone()))
+            .stdin(Box::new(ReadPipe::from(input)))
+            .stderr(Box::new(log_stream.clone()))
+            .build();
         Self {
             wasi,
             wasi_output,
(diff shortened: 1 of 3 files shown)


crates/cli/tests/runner/mod.rs
```diff
@@ -43,12 +43,13 @@ struct StoreContext {
 
 impl StoreContext {
     fn new(input: &[u8], capacity: usize) -> Self {
-        let mut wasi = WasiCtxBuilder::new().inherit_stdio().build();
```
jeffcharles, line 46: I was getting a warning about the `mut` being unnecessary but removing it caused Rust Analyzer to complain that the code had compilation errors even though cargo was able to compile it. I opted to just set the streams in the `WasiCtxBuilder` instead to get around the issue.

jianghong (approved):
Nice! I'm curious if you ran into the issue I had? And what fixed it if so

Issue #310, opened 2023-04-21
New crates to support reusing API implementations


jeffcharles:
## Summary of changes

We'll be:
- adding a `javy` crate that exposes a `Runtime` and `Config` that wraps `JSContextRef` with the ability to access the underlying `JSContextRef` (this may be revisited in a future iteration
- adding an `apis` crate that contains reusable implementations for `console`, `TextEncoder`, `TextDecoder`, `Javy.IO.readSync`, and `Javy.IO.writeSync` that can be added to a `Runtime`

## Motivations

We want to have an easier to understand API for configuring and managing the JS environment used within Javy. We also want other developers building on top of Javy to be able to reuse our implementations of `console`, `TextEncoder`, `TextDecoder`, `Javy.IO.readSync`, and `Javy.IO.writeSync` if they want to and for it to be relatively straightforward.

## Tasks

- [x] Add `javy` crate with `Runtime` and `Config`
- [x] Add `apis` crate with all of the implementations above
- [x] Publish crates to crates.io

jeffcharles:
This has been fully completed! 🎉

PR #330, opened 2023-04-26
More documentation for javy crate

jeffcharles:
Part of #310. Adds a README, some more metadata to the Cargo file, a publish script, and tries to use a more useful code example for the docs.

### DIFF
--- a/crates/javy/Cargo.toml
+++ b/crates/javy/Cargo.toml
@@ -4,7 +4,11 @@ version = "1.0.0-alpha1"
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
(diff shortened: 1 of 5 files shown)


crates/javy/Cargo.toml
```diff
@@ -4,7 +4,11 @@ version = "1.0.0-alpha1"
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
+description = "Configurable JavaScript runtime for WebAssembly"
+homepage = "https://github.com/Shopify/javy/tree/main/crates/javy"
```
jianghong, line 8: Just a note that we'll need to change the URL here later

PR #331, opened 2023-04-26
Add apis crate scaffolding

jeffcharles:
Part of #310. Adds scaffolding for the `javy-apis` crate that will eventually hold `core`'s implementations of console, stream IO, and text encoding APIs.

### DIFF
--- a/crates/apis/src/lib.rs
+++ b/crates/apis/src/lib.rs
@@ -25,6 +25,18 @@
 //! # Ok::<(), Error>(())
 //! ```
 
+#![deny(
+    missing_docs,
+    rustdoc::broken_intra_doc_links,
+    rustdoc::private_intra_doc_links,
+    rustdoc::missing_crate_level_docs,
+    rustdoc::private_doc_tests,
+    rustdoc::invalid_codeblock_attributes,
+    rustdoc::invalid_html_tags,
+    rustdoc::invalid_rust_codeblocks,
+    rustdoc::bare_urls
+)]
+
 use anyhow::Result;
 use javy::Runtime;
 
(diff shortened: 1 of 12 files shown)


jianghong (commented):
Looks good so far

crates/apis/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+use anyhow::Result;
+use api_config::APIConfig;
+use javy::Runtime;
+
+pub use runtime_ext::RuntimeExt;
+
+mod api_config;
+mod runtime_ext;
+
+pub(crate) trait JSApiSet {
```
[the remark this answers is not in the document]
jeffcharles: This is embarrassing, I blanked on doing that. I'll do that this morning.

crates/apis/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+//!            .to_string();
+//!        println!("{str}");
+//!        Ok(JSValue::Undefined)
+//!    })?,
+//! )?;
+//! context.eval_global("hello.js", "print('hello!');")?;
+//! # Ok::<(), Error>(())
+//! ```
+
+#![deny(
```
jeffcharles, line 28: I want to add this at the crate level to trigger failures for certain issues with docs.

We could do this in the Makefile but that means just running a plain `cargo docs` or `cargo check` won't cause them to raise. `missing_docs` being in here should also cause the Rust Analyzer to indicate an error when there are missing docs. I am open to moving it to the Makefile though.

jeffcharles: I'm also open to removing this change from this PR and figuring out an approach we can use consistently across our crates. But I'd like to eventually have CI fail if a public API on the crates we're publishing do not include docs.

saulecabrera: I'd be in favor of removing this change and then figuring out a way to consistently handle docs across crates. My point of view is that we'd need to figure out a consistent way to do this anyway given that we want to add documentation to Javy as whole.

PR #335, opened 2023-05-01
Safer `JSValueRef`

jianghong:
Part of https://github.com/Shopify/javy/issues/295

This PR changes `JSValueRef` to contain a reference to `JSContextRef` instead of a pointer to a quickjs JSContext. 

**Before**
```rust
pub struct JSValueRef {
    pub(super) context: *mut JSContext,
    pub(super) value: JSValue,
}
```

**After**
```rust
pub struct JSValueRef<'a> {
    pub(super) context: &'a JSContextRef,
    pub(super) value: JSValue,
}
```

This change ensures that `JSValueRef` can not outlive the context that it was created from and guarantees safe code. The rest of the changes in the PR follow from having to introduce lifetime annotations. There should be no functional changes.

Also, with this PR, there is now documentation for all the top level API of `quickjs-wasm-rs` 🎉 
image

Next steps would be to put out a 1.0.0 release and publish on crates.io

### DIFF
--- a/crates/quickjs-wasm-rs/src/js_binding/error.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/error.rs
@@ -7,22 +7,17 @@ use std::fmt::{self, Display, Formatter};
 /// This enum provides a convenient way to classify and handle different types of JavaScript errors
 /// that may be encountered. Each variant includes an associated error message to help with
 /// debugging and error reporting.
-///
-/// # Variants
-///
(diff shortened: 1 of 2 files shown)


jeffcharles (approved):
Nit: If you're adding public documentation, you should consider adding crate documentation too (that is, have a series of `//!` comments in `src/lib.rs`). But that could be done in a future PR too.

Otherwise, LGTM 🎉

jianghong:
Added some crate level docs

PR #337, opened 2023-05-01
Tests fuel consumed (but on macos)

jianghong:


jeffcharles (approved):

saulecabrera:
The tests were failing because we were still using the linux version of Wasmtime, so I updated that and also updated the `WASMTIME_VERSION` env variable to `8.0.0` to match the CLI's version.

PR #339, opened 2023-05-01
Introduce cargo vet

jianghong:
Closes https://github.com/Shopify/javy/issues/314

This PR introduces cargo vet and imports some audits from [the trusted registry](https://mozilla.github.io/cargo-vet/importing-audits.html#the-registry).

This was done by:
1. `cargo vet init`
2. Add the imports to `config.toml`
3. Run `cargo vet prune`. This marks the audited crates from the imported list as "audited".

As a start, I chose to import Bytecode Alliance's audit as well as the [imports that wasmtime currently import](https://github.com/bytecodealliance/wasmtime/blob/d5d2e854f97daa3acc6820299231e73cbb0b246c/supply-chain/config.toml#L7-L20).

With these imports, here are the stats for `cargo vet`
```
Shopify/javy [jh/cargo-vet]$ cargo vet
Vetting Succeeded (83 fully audited, 1 partially audited, 201 exempted)
```

### DIFF
--- a/.github/workflows/cargo-vet.yml
+++ b/.github/workflows/cargo-vet.yml
@@ -1,0 +1,27 @@
+name: Cargo vet dependencies
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    env:
+      CARGO_VET_VERSION: 0.6.1
+    steps:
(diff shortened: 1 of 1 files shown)


saulecabrera (approved):
Changes LGTM, do you think we should add a step on CI to run cargo vet?

jianghong:
I think that makes sense. I'll have to play around with what is considered a fail state for the CI first

jeffcharles (approved):

PR #340, opened 2023-05-02
quickjs-wasm-rs v1.0.0

jianghong:
Closes https://github.com/Shopify/javy/issues/295

Bump the version to v1.0.0 and add a changelog.

If this looks good, I'll publish to crates.io after merging.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1444,7 +1444,7 @@ dependencies = [
 
 [[package]]
 name = "quickjs-wasm-rs"
-version = "1.0.0-alpha.3"
+version = "1.0.0"
 dependencies = [
  "anyhow",
  "once_cell",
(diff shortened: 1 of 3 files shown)


saulecabrera (approved):

jeffcharles:
We still have docs missing on a few public members, for example, `JSContextRef`'s `eval_global`. Is it in scope to add documentation for those members before releasing 1.0.0 or is that something we want to do later?

jianghong:
Oh yeah good point. I wonder if this is a good time to decrease our API surface by making a bunch of the methods on `JSContextRef` and `JSValueRef` private, or deleting them altogether to encourage the use of `JSValue`? Like `eval_global` could return `JSValue`. Any concerns with that? 🤔

jeffcharles:
We should also move the `json` and `messagepack` modules out to the `javy` crate before releasing this as 1.0.

PR #343, opened 2023-05-02
Add TextCodec API to support TextEncoder and TextDecoder

jianghong:
Part of https://github.com/Shopify/javy/issues/310

Add the `TextCodec` API which supports `TextEncoder` and `TextDecoder`. 

The implementation is exactly how we currently do that - that is, it is backed by some JavaScript code as well. I opted to create the `text_codec` module as a directory that contains `mod.rs` and the `text-encoding.js` file. I think this pattern is nice as it packages the rust and js files in 1 directory instead of it being at the top level.


crates/apis/src/text_codec/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    #[test]
+    fn test_text_encoder_decoder() -> Result<()> {
+        let runtime = Runtime::default();
+        let context = runtime.context();
+        let text_codec_api = TextCodec;
+        text_codec_api.register(&runtime, &APIConfig::default())?;
+        let result = context.eval_global(
+            "main",
+            "let encoder = new TextEncoder(); let buffer = encoder.encode('hello'); let decoder = new TextDecoder(); decoder.decode(buffer) == 'hello';"
```
jianghong, line 98: Opted to test this by evaluating this statement. We used to this this by printing to a in memory steam and asserting on that, but we shouldn't assume the `Console` api is available in this api, so I think this is the next best way

jianghong:
Going to wait until https://github.com/Shopify/javy/pull/345/files is merged and then I'll follow on with the same patterns that the rest of the PR has

jianghong:
Please reopen

PR #345, opened 2023-05-02
Move console API to apis crate

jeffcharles:
Part of #310. Moves the `console` API from `javy-core` to `javy-apis`.

### DIFF
--- a/crates/apis/src/api_config.rs
+++ b/crates/apis/src/api_config.rs
@@ -1,3 +1,64 @@
+use std::io::{self, Write};
+
+#[cfg(feature = "console")]
+/// A selection of possible destination streams for `console.log` and
+/// `console.error`.
+#[derive(Debug)]
+pub enum LogStream {
+    /// The standard output stream.
+    StdOut,
+    /// The standard error stream.
+    StdErr,
+}
+
+#[cfg(feature = "console")]
+impl LogStream {
+    pub(crate) fn as_stream(&self) -> Box<dyn Write + 'static> {
+        match self {
+            Self::StdErr => Box::new(io::stderr()),
+            Self::StdOut => Box::new(io::stdout()),
+        }
+    }
+}
+
 /// A configuration for APIs added in this crate.
-#[derive(Debug, Default)]
-pub struct APIConfig {}
+///
+/// Example usage:
+/// ```
+/// # use javy_apis::{APIConfig, LogStream};
+/// let api_config = APIConfig::default();
+/// ```
+#[derive(Debug)]
+pub struct APIConfig {
+    #[cfg(feature = "console")]
+    pub(crate) log_stream: LogStream,
+    #[cfg(feature = "console")]
+    pub(crate) error_stream: LogStream,
+}
+
+impl APIConfig {
+    #[cfg(feature = "console")]
+    /// Sets the destination stream for `console.log`.
+    pub fn log_stream(&mut self, stream: LogStream) -> &mut Self {
+        self.log_stream = stream;
+        self
+    }
+
+    #[cfg(feature = "console")]
+    /// Sets the destination stream for `console.error`.
+    pub fn error_stream(&mut self, stream: LogStream) -> &mut Self {
+        self.error_stream = stream;
+        self
+    }
+}
+
+impl Default for APIConfig {
+    fn default() -> Self {
+        Self {
+            #[cfg(feature = "console")]
+            log_stream: LogStream::StdOut,
+            #[cfg(feature = "console")]
+            error_stream: LogStream::StdErr,
+        }
+    }
+}
--- a/crates/apis/src/api_config.rs
+++ b/crates/apis/src/api_config.rs
@@ -25,7 +25,7 @@ impl LogStream {
 ///
 /// Example usage:
 /// ```
-/// # use javy_apis::{APIConfig, LogStream};
+/// # use javy_apis::APIConfig;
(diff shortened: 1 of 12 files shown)


crates/core/src/globals.rs
```diff
@@ -172,113 +133,18 @@ mod tests {
...
         let ctx = runtime.context();
-        inject_javy_globals(&runtime, stream.clone(), stream.clone())?;
+        inject_javy_globals(&runtime)?;
         ctx.eval_global(
             "main",
-            "let encoder = new TextEncoder(); let buffer = encoder.encode('hello'); let decoder = new TextDecoder(); console.log(decoder.decode(buffer));"
+            "let encoder = new TextEncoder(); let buffer = encoder.encode('hello'); let decoder = new TextDecoder(); globalThis.foo = decoder.decode(buffer);"
         )?;
-        assert_eq!(b"hello\n", stream.buffer.borrow().as_slice());
+        assert_eq!("hello", ctx.global_object()?.get_property("foo")?.as_str()?);
```
jeffcharles, line 146: We no longer can pass an in-memory stream into the `console` implementation provided by the `javy-apis` crate so I rewrote this test to write the decoded value to a global variable which we can inspect directly.

crates/apis/src/api_config.rs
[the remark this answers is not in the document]
jianghong: Yeah I was getting the same feeling of how it could get messy. One thing I thought of that could reduce it a little bit is to group the `APIConfig` fields per API? So in this case it would be 

```rust
struct ConsoleConfig {
  log_stream: ..
  error_stream: ..
}
struct APIConfig
  #[cfg(feature = "console")]
  console: ConsoleConfig,
}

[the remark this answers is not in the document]
Comment (jianghong) [reply crates/apis/src/api_config.rs:1]: I don't think that's as important. My gut feeling is that 99% of the config options will be pretty specific to each API. That's just an assumption.

Maybe it would be helpful to code up this approach to see how it looks in terms of simplicity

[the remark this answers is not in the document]
Comment (jeffcharles) [reply crates/apis/src/api_config.rs:1]: I'm very confident what you've proposed will be simpler.

Having the `ConsoleConfig` in the in the `Console` module will introduce a circular dependency between the `api_config` and `console` modules. That's okay within a crate but a circular dependency makes it more difficult to split out `console` into its own crate later and that was something we were talking about doing. But, I suppose we can deal with that later.

I can make the change you've proposed.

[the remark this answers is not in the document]
Comment (jeffcharles) [reply crates/apis/src/api_config.rs:1]: I actually figured out a slightly more clever thing we can do where we can define some setter methods on `APIConfig` in the `Console` module so we don't even need to expose that there's an underlying `ConsoleConfig` struct at all. It looks really clean IMO.

File: crates/apis/src/console/config.rs [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+        Self {
+            log_stream: LogStream::StdOut,
+            error_stream: LogStream::StdErr,
+        }
+    }
+}
+
+impl APIConfig {
+    /// Sets the destination stream for `console.log`.
+    pub fn log_stream(&mut self, stream: LogStream) -> &mut Self {
```
jianghong, line 41: What happens if another API happens to implement the same name `log_stream` here?

jianghong: Just to expand on my comment and concern here:

I'm wondering if this pattern would mean that API creators will have to be aware of the possible names of other API configs so they don't clobber. That might be a burden we'd want to avoid

jianghong: Right, that's good that there's a warning at least. But I assume this warning would also only appear if they have the conflicting feature turned on. IMO it feels like we'd want to avoid this issue altogether. What do you think?

jeffcharles: > I'm wondering if this pattern would mean that API creators will have to be aware of the possible names of other API configs so they don't clobber.

I don't think API creators would have to be aware of all possible names. The compiler will let them know if they clobber if all features are turned on. They can also look at the docs for the crate or some kind of autocompletion to get a full list of methods on `APIConfig` if they clobber and want to ensure the next name they try isn't already defined on `APIConfig`.

jianghong: For sure, I guess I don't like that they are limited by the names they can pick here, especially if these will end up being separate crates eventually.

If you feel that's not that big of a deal, I'm okay with approving this

jianghong (approved):

crates/apis/src/console/config.rs
saulecabrera: Just catching up on this. I do like the approach followed in this PR in order to ease the `cfg` complexity (e.g. having dedicated structs/enums for each API configuration).

On the other hand, my gut tells me that we could start simple, by following the second approach described in the previous comment

```rust
config.log_stream(...)
```


The reason for this preference is mostly to stick to the principle that configuration should happen via a single, top-level entry point, with the hope to make it easier to reason about. In the first approach we'd almost be configuring each API separately, which risks simplicity.

All that said, this is something that we can keep an eye on, and change if there's need to. I do agree though, that consumers will be limited by the names that they can pick for each API, but hopefully this won't happen often.

jeffcharles: I'm going to go ahead and merge this then since it uses the second approach. We do use sub-structs behind the scenes but the consumer facing API for `APIConfig` hides all of that.

PR #368, opened 2023-05-16
quickjs-wasm-sys v1.0.0

jeffcharles:
Cutting a v1.0.0 for `quickjs-wasm-sys` since we're confident in the bindings APIs.


saulecabrera (approved):
Changes LGTM, you'd need to [record an audit](https://mozilla.github.io/cargo-vet/recording-audits.html) for this version and I think this would be ready to go

PR #376, opened 2023-05-18
Update cargo-vet settings

jeffcharles:
Let me know if you're not comfortable with marking our packages as audited.

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -1,6 +1,21 @@
 
 # cargo-vet audits file
 
+[[audits.javy]]
+who = "Jeff Charles <jeff.charles@shopify.com>"
+criteria = "safe-to-deploy"
+version = "1.0.0"
+
+[[audits.javy-apis]]
+who = "Jeff Charles <jeff.charles@shopify.com>"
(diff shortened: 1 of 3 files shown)


saulecabrera (approved):

saulecabrera:
> Let me know if you're not comfortable with marking our packages as audited.

All this looks reasonable to me!

jeffcharles:
I'm actually realizing `audit-as-crates-io` should probably be set to `false` because vetting will break as soon as a version change happens

jeffcharles:
For context, I read through https://mozilla.github.io/cargo-vet/first-party-code.html. My understanding is that `audit-as-crates-io` exists to handle cases where your patching a third-party crate in your project and should be set to `true` when the crate contains third-party code. I don't think that applies in our case so it should be set to `false`. Though we can still release audits for our own crates of course.

saulecabrera:
Good catch, reading through the description that is my understanding as well, I still have some open questions, but I don't think we need to block this PR on those, and perhaps we can take them offline.

saulecabrera (approved):

Issue #317, opened 2023-04-25
Unexpected non-deterministic output when recompiling Javy provider


jianghong:
Given a JS file that has the following source code:

```javascript
console.log(Math.random())
```

I see the following results:

```
➜  javy git:(main) target/release/javy compile -d index.js -o index.wasm                                                                   
➜  javy git:(main) wasmtime run --preload javy_quickjs_provider_v1=target/wasm32-wasi/release/javy_quickjs_provider_wizened.wasm index.wasm
0.6111985957948094
➜  javy git:(main) wasmtime run --preload javy_quickjs_provider_v1=target/wasm32-wasi/release/javy_quickjs_provider_wizened.wasm index.wasm
0.6111985957948094
➜  javy git:(main) make core
cargo build --package=javy-core --release --target=wasm32-wasi --features=
    Finished release [optimized] target(s) in 0.09s
wizer target/wasm32-wasi/release/javy_quickjs_provider.wasm --allow-wasi --wasm-bulk-memory true -o target/wasm32-wasi/release/javy_quickjs_provider_wizened.wasm
➜  javy git:(main) wasmtime run --preload javy_quickjs_provider_v1=target/wasm32-wasi/release/javy_quickjs_provider_wizened.wasm index.wasm
0.3228588863524102
➜  javy git:(main) wasmtime run --preload javy_quickjs_provider_v1=target/wasm32-wasi/release/javy_quickjs_provider_wizened.wasm index.wasm
0.3228588863524102
```

Running the same module with the same provider twice has identical output. This is expected for Shopify Functions. I've filed https://github.com/Shopify/javy/issues/318 for the non-Shopify use case since this is unexpected for that. However, simply recompiling the provider and linking to the recompiled provider, and then re-running the same module results in a different output. If there were a code change in the provider code, this would be somewhat expected. But there is no change which makes this unexpected behaviour to me and possibly for a Shopify Functions developer.

An additional impact of this bug is that our builds are non-deterministic. Rebuilding with the same code will always yield a slightly different executable which makes investigating some issues more difficult.

jeffcharles:
The root cause of this issue and #318 is that a `random_state` property on the JS context struct is set according to the current time during JS context initialization and we run that initialization as part of the Wizer preinitialization steps. That `random_state` is used as a seed by `Math.random()`.

[...]

PR #377, opened 2023-05-23
Initialize random seed on first use of random

### LINKED ISSUE #318: Random seems broken in Javy
Given a JS source file that looks like:

```javascript
console.log(Math.random())
```

I see the following output:

```
➜  javy git:(main) target/release/javy compile index.js -o index.wasm                                                                   
➜  javy git:(main) wasmtime run index.wasm
0.6111985957948094
➜  javy git:(main) wasmtime run index.wasm
0.6111985957948094
```

Running the same module with a random call multiple times results in the same output. Random should not be deterministic when not using a `deterministic-wasi-ctx`.

jeffcharles:
Fixes #318 and #317. Technically not all of #317 in that the build output is still non-deterministic but the output of `Math.random` will be either deterministic or non-deterministic based on the WASI context used at runtime instead of based on the WASI context used at compile time.

### DIFF
--- a/crates/apis/src/random/mod.rs
+++ b/crates/apis/src/random/mod.rs
@@ -1,0 +1,37 @@
+use anyhow::Result;
+use javy::{quickjs::JSValue, Runtime};
+
+use crate::{APIConfig, JSApiSet};
+
+pub struct Random;
+
+impl JSApiSet for Random {
+    fn register(&self, runtime: &Runtime, _config: &APIConfig) -> Result<()> {
+        let ctx = runtime.context();
+        ctx.global_object()?.get_property("Math")?.set_property(
+            "random",
+            // TODO figure out if we can lazily initialize the PRNG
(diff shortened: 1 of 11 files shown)


crates/apis/src/random/mod.rs
saulecabrera, line 1: I'm still wondering if we should add module-level documentation explaining when you should use this feature and what are the consequences of doing so? As a consumer I might be interested in understanding: 

* Any module size implications
* Any performance implications

Both of those compared to the default implementation.

My hope is that with clear documentation we'd be able to ensure that consumers understand when they should not pick the default, as originally stated in https://github.com/bytecodealliance/javy/issues/317#issuecomment-1551165567

jeffcharles: I added some additional comments in both the module level docs and the README. Let me know what you think.

saulecabrera (approved):

Issue #360, opened 2023-05-12
Gracefully handle JS execution errors


saulecabrera:
## What is the idea?

At the moment, we are calling `unwrap` on the result of executing JavaScript (e.g. https://github.com/Shopify/javy/blob/main/crates/core/src/lib.rs#L67), we should gracefully handle the error case of the result. 

This is also the case in https://github.com/Shopify/javy/blob/main/crates/core/src/main.rs#L32

## What problem does it solve?

Reduce the noise generated by Rust when unwrapping an error value and provide better feedback when JS execution fails.

nickwesselman:
Ideally we would output the JavaScript stack trace as part of this, assuming it's available in QuickJS.

PR #378, opened 2023-05-24
Handle JS errors gracefully

Maaarcocr:
# why? 

should fix https://github.com/bytecodealliance/javy/issues/360

# how? 

instead of plain unwrapping, handle the errors on both call sites for `run_bytecode` and print out the error if it's there. This prints also the stacktrace from JS, which was already been captured by existing code. 

# question for the reviewers

should we still panic after we have handled the error?

### DIFF
--- a/crates/cli/tests/dylib_test.rs
+++ b/crates/cli/tests/dylib_test.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use std::boxed::Box;
 use std::str;
-use wasi_common::pipe::WritePipe;
+use wasi_common::{pipe::WritePipe, WasiFile};
 use wasmtime::{Engine, Instance, Linker, Store};
 use wasmtime_wasi::{sync::WasiCtxBuilder, WasiCtx};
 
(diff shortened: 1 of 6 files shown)


crates/core/src/lib.rs
```diff
@@ -64,7 +64,12 @@ pub unsafe extern "C" fn compile_src(js_src_ptr: *const u8, js_src_len: usize) -
 pub unsafe extern "C" fn eval_bytecode(bytecode_ptr: *const u8, bytecode_len: usize) {
     let runtime = RUNTIME.get().unwrap();
     let bytecode = slice::from_raw_parts(bytecode_ptr, bytecode_len);
-    execution::run_bytecode(runtime, bytecode).unwrap();
+    match execution::run_bytecode(runtime, bytecode) {
+        Ok(_) => {}
+        Err(e) => {
+            eprintln!("Error while running JS: {}", e);
+        }
```
jeffcharles, line 71: We do want the error case to trigger a trap since we want code that has failed to run to signal that it has failed to run to whatever is invoking it. However, we don't want a panic because that adds a lot of noise to the logs that we do not want. Using `std::process::abort()` is one option. Another is invoking `std::process::exit` with the canonical `FAILURE` `std::process::ExitCode` value.

Also is there a reason not to move this logic into the `execution` module given the code here and in `main.rs` will be identical?

Final nit, you can simplify the `match` statement to an `or_else` statement given the `Ok` branch is not performing any operations.

crates/cli/tests/dylib_test.rs
```diff
@@ -36,6 +36,36 @@ fn test_dylib() -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_dylib_with_error() -> Result<()> {
```
jeffcharles, line 40: This new test function shares a lot of code with the function above, can it be deduplicated?

jeffcharles (changes requested):
Thank you for the contribution!

A couple things that need to be addressed:
- A failure to evaluate the bytecode needs to trigger a trap without logging a panic
- A test needs to be added to validate that statically compiled modules that trigger an error display the expected behaviour. The test added only validates this for dynamic modules. As well, the tests will need to validate that a trap occurs.

Maaarcocr:
Thank you for the super insightful comments. Will address those!

jeffcharles (approved):
LGTM!

Thank you again for your contribution!

jeffcharles (approved):
LGTM!

Thank you again for your contribution!

PR #383, opened 2023-05-30
Download WASI SDK & run Wizer via `build.rs`

surma:
This PR proposes and implements two changes:
- Automatically download the correct WASI SDK in `quickjs-wasm-sys`’s `build.rs` (can still be overwritten via the `QUICKJS_WASM_SYS_WASI_SDK_PATH` env variable)
- Run Wizer in `javy-cli`’s `build.rs`

Why:
- Remove the somewhat unidiomatic use of a Makefile
- Hopefully will also fix [docs.rs](https://docs.rs/crate/quickjs-wasm-rs) for these crates
- Build instruction are less opaque/mysterious (see changes to README)


crates/quickjs-wasm-sys/build.rs
```diff
@@ -1,21 +1,74 @@
...
+        let file_suffix = match (env::consts::OS, env::consts::ARCH) {
+            ("linux", "x86") | ("linux", "x86_64") => "linux",
+            ("macos", "x86") | ("macos", "x86_64") | ("macos", "aarch64") => "macos",
+            ("windows", "x86") => "mingw-x86",
+            ("windows", "x86_64") => "mingw",
+            other => return Err(anyhow!("Unsupported platform tuple {:?}", other)),
+        };
+
+        let mut archive = fs::File::create(&archive_path)?;
+        reqwest::blocking::get(&format!("https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-19/wasi-sdk-19.0-{}.tar.gz", file_suffix))?.copy_to(&mut archive)?;
```
jeffcharles, line 27: Can we extract the version numbers into constants (or variables) so it's easier to update them?

crates/quickjs-wasm-sys/build.rs
```diff
@@ -1,21 +1,74 @@
...
+    }
+
+    let mut test_binary = wasi_sdk_dir.clone();
+    test_binary.extend(["bin", "wasm-ld"]);
+    // Extract archive if necessary
+    if !test_binary.try_exists()? {
+        let output = process::Command::new("tar")
```
jeffcharles, line 34: I suspect this would fail on Windows if not running in WSL. That's probably fine given we don't currently support have great support for this outside of WSL (though we don't document that). But what would the output look like if `tar` isn't on the `PATH`? I just want to make sure it's not too confusing for a Windows user.

jeffcharles (commented):
Looks good to me, just a couple questions

crates/quickjs-wasm-sys/build.rs
surma: It should error with the Windows equivalent of `Command not found: tar`.

crates/quickjs-wasm-sys/build.rs
jeffcharles, line 27: Also probably should bump the version for the WASI SDK to 20 given your previous PR

crates/quickjs-wasm-sys/build.rs
```diff
@@ -1,21 +1,77 @@
...
+        let file_suffix = match (env::consts::OS, env::consts::ARCH) {
+            ("linux", "x86") | ("linux", "x86_64") => "linux",
+            ("macos", "x86") | ("macos", "x86_64") | ("macos", "aarch64") => "macos",
+            ("windows", "x86") => "mingw-x86",
+            ("windows", "x86_64") => "mingw",
+            other => return Err(anyhow!("Unsupported platform tuple {:?}", other)),
+        };
+
+        let mut archive = fs::File::create(&archive_path)?;
+        reqwest::blocking::get(format!("https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-{}/wasi-sdk-{}.{}-{}.tar.gz", WASI_SDK_VERSION_MAJOR, WASI_SDK_VERSION_MAJOR, WASI_SDK_VERSION_MINOR, file_suffix))?.copy_to(&mut archive)?;
```
jeffcharles, line 30: Nit: I _think_ you can put the variables inside the `{}`s which would make it a little easier to read. See https://doc.rust-lang.org/std/fmt/#usage.

Please do test that this compiles since I haven't tried it on my machine yet:
```suggestion
        reqwest::blocking::get(format!("https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-{WASI_SDK_VERSION_MAJOR}/wasi-sdk-{WASI_SDK_VERSION_MAJOR}.{WASI_SDK_VERSION_MINOR}-{file_suffix}.tar.gz"))?.copy_to(&mut archive)?;
```

jeffcharles (approved):

surma:
New approach:
- Use `hyper` (and therefore `tokio`) as a HTTP client
- Copy [the hyper-related and tokio-related exemptions](https://github.com/bytecodealliance/wasmtime/blob/c044790a9047db1593516b9c6916642c155ee87a/supply-chain/config.toml#L322-L325) that the wasmtime team is using
- Add additional exemptions for `openssl` and `openssl-sys` so we can create TLS connections (GitHub enforces HTTPS)
- Pin all the transitive dependencies exactly to the exempted versions in `build-dependencies`

crates/cli/Cargo.toml
```diff
@@ -37,6 +37,10 @@ criterion = "0.3"
 num-format = "0.4.3"
 tempfile = "3.4.0"
 
+[build-dependencies]
+anyhow = "1.0.71"
+wizer = { git = "https://github.com/bytecodealliance/wizer" }
```
jeffcharles, line 42: Can you rebase on main and set the `rev` for this? We realized yesterday that `cargo install` ignores the `Cargo.lock` file so we can get into a state where `cargo install` fails when `main` moves far enough ahead of what's in the lockfile.

saulecabrera (approved):

jeffcharles (approved):

Issue #390, opened 2023-06-08
Error while Building


Thorngage:
Hello, I'm a complete Noob trying to test out Javy to compile a super simple .js file to wasm. 
I'm doing this research for a team of much more experienced people than me that, however, are not versed enough in english to do this research themselves (the almost totality of documentation regarding Javy is in english), so really forgive me if whatever I'm doing is obviously wrong, but most of what I've been doing, I did not know what it was.

So I'm working on a windows machine, after installing with much difficulty the WSL (using Ubuntu as a distro)
I've been following the commands in the readMe documentation (could use some more details, the need to install rust is not clear to a newbie, as are some things that don't work without installing others first, the rustup and cargo commands did not work at first, but fortunately Linux's terminal is very helpful in identifying problems)
The first big hurdle was the error linker 'cc' not found, I found out from some other website that APT update was necessary, so maybe adding a mention to update and eventually run 

`sudo apt-get update`
and
`sudo apt install build-essential`
before installing wasmtime-cli could help other newbies like me

After that I kept getting an error message of not finding the Cargo.toml file in the directory in the first building step, however after a while I realized that the command
`cargo build -p javy-core --target=wasm32-wasi -r`
was to be run inside the javy directory (was I right?)

I ran the command there,  it seemed to go all right but now i get an extensive error message, apparently i needed to instal a pkg-config

i tried running
`apt install pkg-config`
but then when I re-run the command inside javy i get this

```
  Compiling openssl-sys v0.9.88
   Compiling openssl-macros v0.1.0
   Compiling tokio-macros v1.7.0
   Compiling minimal-lexical v0.2.1
   Compiling httparse v1.8.0
   Compiling once_cell v1.16.0
   Compiling native-tls v0.2.11
   Compiling futures-channel v0.3.27
error: failed to run custom build command for `openssl-sys v0.9.88`

Caused by:
  process didn't exit successfully: `/home/michele/javy/target/release/build/openssl-sys-ca1c01bffa9b51a6/build-script-main` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR
  X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR unset
  cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
  OPENSSL_LIB_DIR unset
[...]

PR #393, opened 2023-06-09
Try to clarify parts of README including instructions for Ubuntu

jeffcharles:
Addressing some points in #390. The requirement for the `clang` package in particular was a little tricky to figure out.

I'd also like to move the instructions for how to build Javy below the instructions for how to use Javy now that we have precompiled binaries but I'll do that in a separate PR to keep the diff easier to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,11 @@ $ cargo build -p javy-core --target=wasm32-wasi -r
 $ cargo build -p javy-cli -r
 ```
 
-Alternatively you can run `cargo install --path crates/cli` from the root of the Javy repository to install the Javy CLI globally.
+Alternatively if you want to install the Javy CLI globally, inside the Javy repository run:
+```
+$ cargo build -p javy-core --target=wasm32-wasi -r
+$ cargo install --path crates/cli
+```
 
 ## Using Javy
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,35 +27,41 @@ We welcome feedback, bug reports and bug fixes. We're also happy to discuss feat
...
+## How to build
 
+Inside the Javy repository, run:
 ```
 $ cargo build -p javy-core --target=wasm32-wasi -r
 $ cargo build -p javy-cli -r
 ```
 
-Alternatively you can run `cargo install --path crates/cli` to install the Javy CLI globally.
+Alternatively you can run `cargo install --path crates/cli` from the root of the Javy repository to install the Javy CLI globally.
```
surma, line 58: One idea that I realized wasn’t clear before either: The command listed here is an alternative for the second command, not both.

```suggestion
Alternatively you can run `cargo install --path crates/cli` from the root of the Javy repository after building `javy-core` to install the Javy CLI globally.
```

surma (approved):

README.md
jeffcharles: Good catch! I opted to go with a slightly different way of phrasing that to try and make the exact instructions and sequence even more explicit.

Issue #394, opened 2023-06-13
Does `CallbackArg` still make sense?


TheEdward162:
Hi.

As far as I can tell `CallbackArg` was added in https://github.com/bytecodealliance/javy/pull/309 so that there would be a lifetime on `wrap_callback` this and args parameters and they wouldn't have to get eagerly resolved to `JSValue`. However, a week later https://github.com/bytecodealliance/javy/pull/335 introduced the same lifetime on `JSValueRef`, which leads me to believe the `CallbackArg` struct is obsolete now.

This is painful to deal with because I have some existing functions which I hook into callbacks from JS similar to this:
```rs
fn higher_level_callback<H: MyStateTrait>(
    state: &mut H,
    context: &Context,
    _this: JsValue,
    args: &[JsValue],
) -> Result<JsValue, JSError> {
    let response = match args.get(0) {
         None => return Err(JSError::Type("some error".to_string())),
         Some(name) => format!("Hello {}", name)
    }

    Ok(context.value_from_str(&response).unwrap())
}

// then somewhere in a macro I set up the callback and set it on an object, which I later inject into the global context
{
    let state = $state.clone();
    let fun = $context.wrap_callback(move |context, this, args| {
        let result = $fn_impl(state.borrow_mut().deref_mut(), context, this, args).map_err(anyhow::Error::from); // this should also not be anyhow, but that's another issue
        // some logging
        result
    })?;
    $parent.set_property($key, fun)?;
}
```

Converting this to a newer approach leads to this:
```rs
fn __export_message_exchange<'ctx, H: MyStateTrait>(
    state: &mut H,
    _this: JSValueRef<'ctx>,
    args: &[JSValueRef<'ctx>],
) -> Result<JSValue, JSError> {
    let response = match args.get(0) {
         None => return Err(JSError::Type("some error".to_string())),
         Some(name) => format!("Hello {}", name)
    }

    Ok(JSValue::String(response))
}

// the macro
{
    let state = $state.clone();
    let fun = $context.wrap_callback(move |context, this, args| {
        let this = unsafe { this.inner_value() }; // this isn't unsaf, and is just a noop anyway
        let args: Vec<_> = unsafe { args.into_iter().map(|a| a.inner_value()).collect() }; // this map->collect is pointless
        let result = $fn_impl(state.borrow_mut().deref_mut(), this, args).map_err(anyhow::Error::from);
        // some logging
        result
    })?;
    $parent.set_property($key, fun)?;
}
```
[...]

PR #395, opened 2023-06-14
remove CallbackArg, use JSValueRef instead

TheEdward162:
As discussed in https://github.com/bytecodealliance/javy/issues/394 this removes CallbackArg and uses the underlying JSValueRef to reduce API friction.

This is a breaking change so it will probably need a major version bump on release.


crates/quickjs-wasm-rs/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Callback functions now work with `JSContextRef` again and the `CallbackArg::try_into` conversions are implemented for it instead.
```
jeffcharles, line 11: We also need an entry added to the changed section for the signature of `from_qjs_value` changing since it now takes ownership of the `JSValueRef` instead of a reference.

crates/quickjs-wasm-rs/src/js_value/qjs_convert.rs
```diff
@@ -22,10 +22,10 @@ use crate::js_binding::{context::JSContextRef, value::JSValueRef};
 ///
 /// ```
 /// // Assuming `qjs_val` is a `JSValueRef` pointing to a QuickJS string
-/// let js_val = from_qjs_value(&qjs_val).unwrap();
+/// let js_val = from_qjs_value(qjs_val).unwrap();
 /// assert_eq!(js_val, "hello".into());
 /// ```
-pub fn from_qjs_value(val: &JSValueRef) -> Result<JSValue> {
+pub fn from_qjs_value(val: JSValueRef) -> Result<JSValue> {
```
jeffcharles, line 28: Can you update the comment on line 11 to indicate it converts the ref and not a reference to it?

crates/quickjs-wasm-rs/src/js_binding/context.rs
```diff
@@ -313,24 +313,24 @@ impl JSContextRef {
 
     /// Wrap the specified function in a JS function.
     ///
-    /// Since the callback signature accepts parameters as high-level `JSContextRef` and `CallbackArg` objects, it can be
+    /// Since the callback signature accepts parameters as high-level `JSContextRef` objects, it can be
```
jeffcharles, line 316: So this also accepts parameters as `JSValueRef` objects.

crates/quickjs-wasm-rs/CHANGELOG.md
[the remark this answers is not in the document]
TheEdward162: Ah, yes, my bad, I mixed up `JSContextRef` and `JSValueRef` at some point.

jeffcharles (approved):
LGTM! Thanks again!

PR #400, opened 2023-06-16
Remove flaky fuel assert for random test case

jeffcharles:
Our CI suite occasionally fails due to the fuel consumed being inconsistent from test run to test run so let's not do this check for `random`.


saulecabrera (approved):
LGTM -- left one minor note.

PR #402, opened 2023-06-19
Refactor to use Walrus instead of wasm-encoder

### LINKED ISSUE #399: Migrating from wasm-encoder to walrus
## What is the idea?

I'll be migrating Javy off of [the wasm-encoder crate](https://crates.io/crates/wasm-encoder) to use [Walrus](https://crates.io/crates/walrus) instead.

We're going to be making some changes to how we generate Wasm modules that involve emitting more raw WebAssembly and the abstractions in Walrus will save me time and make the code easier to maintain.

jeffcharles:
Closes #399.

I had to make the tests around the producers custom section more strict because Walrus adds itself to the producers by default and I wanted to detect if Walrus was doing that and fail the test if it was.

### DIFF
--- a/crates/cli/tests/common/mod.rs
+++ b/crates/cli/tests/common/mod.rs
@@ -1,7 +1,9 @@
 use anyhow::Result;
 use std::path::{Path, PathBuf};
+use wasmparser::ProducersSectionReader;
 use wasmtime::{Engine, Module};
 
+#[allow(dead_code)]
 pub fn create_quickjs_provider_module(engine: &Engine) -> Result<Module> {
     let mut lib_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     lib_path.pop();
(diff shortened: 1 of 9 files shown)


crates/cli/tests/common/mod.rs
```diff
@@ -1,7 +1,9 @@
 use anyhow::Result;
 use std::path::{Path, PathBuf};
+use wasmparser::ProducersSectionReader;
 use wasmtime::{Engine, Module};
 
+#[allow(dead_code)]
```
jeffcharles, line 6: I had to add this because each integration test suite is considered its own application and we use `create_quickjs_provider` module in two of three suites and `assert_producers_section_is_correct` in two of three suites.

jeffcharles:
I had to add Walrus and its immediate dependencies as exemptions since I wasn't able to find any existing audits for Walrus. Gimli has audits for deltas for a future version but not for the version that Walrus depends on. I'm not really sure how else to satisfy the vetting and still use Walrus.

surma (approved):
Nice!

crates/cli/tests/common/mod.rs
saulecabrera: Can we add this comment in the code to give context to readers why that directive exists?

PR #407, opened 2023-06-23
Refactor to add abstraction around the JS source

jeffcharles:
We do a number of operations on JavaScript source code. We compress it, we compile it to QuickJS bytecode, we load it from a file. Soon we're going to be deriving a list of exported functions from it. Based on the work I'm doing for multiple exports, I think now is a good time to treat the JavaScript as its own type which exposes various representations we can derive from it.

I'm open to suggestions if there's a better name than `JS` for the module or struct.

### DIFF
--- a/crates/cli/src/js.rs
+++ b/crates/cli/src/js.rs
@@ -1,3 +1,8 @@
+/// Higher-level representation of JavaScript.
+///
+/// This is intended to be used to derive different representations of source
+/// code. For example, as a byte array, a string, QuickJS bytecode, compressed
+/// bytes, or attributes of the source code like what it exports.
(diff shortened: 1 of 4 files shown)


crates/cli/src/js.rs
saulecabrera, line 1: I think JS is fine as name, but can we add some module level documentation? What I'd like to address through such documentation is "what goes in this module vs what doesn't".

saulecabrera (approved):

crates/cli/src/js.rs
jeffcharles: I made an attempt at describing how it's intended to be used for deriving various representations of the source code. We can iterate on it.

Issue #405, opened 2023-06-21
Add support for exposing exported JS functions as Wasm exports


jeffcharles:
## What is the idea?

When there are exported JS functions with no arguments and no return value, Javy should add exported an exported Wasm function for each exported JS function that will invoke the JS function after evaluating the top-level code in the JS module.

## What problem does it solve?

We have a use case where we want multiple entrypoints into a module produced by Javy.

## Proposed solution

Use SWC's ECMAScript parser to identify functions exported from a JS module and for each exported JS function, using Walrus, add an export to a function implementation that will load the name of the JS function export and then call an invoke function in the Wasm module which will trigger top-level evaluation and execute the exported JS function.

PR #409, opened 2023-06-26
Refactor Wasm generation in CLI

jeffcharles:
After exploring how to introduce changes for #405, I've found our existing CLI crate abstractions would result in the changes looking a little messy. So I wanted to try to sharpen our existing abstractions.

The first big change is in how we build static modules. At the moment, we have:
1. Start a subprocess and set the stdin to the JS
2. In the subprocess, we create and use a struct called `Optimizer` to generate the module from javy-core's bin output with Wizer and optionally run it through `wasm-opt` and then write it to disk
3. We continue in the parent process, load the output file from disk, run a Walrus transform to add two custom sections, and then write the result to disk

The change is:
1. Start a subprocess and set the stdin to the JS
2. In the subprocess, we create and use a static module `wasm_generator` which loads javy-core's bin output and processes it with Wizer and then write the result to disk
3. We continue in the parent process, load the output file from disk, and we create and use a static module `Refiner` to optionally optimize the Wasm and use Walrus to add two custom sections, and then write the result to disk.

It's the same set of actions occurring except we have generation that happens in the subprocess happen in a `Generator` abstraction and the steps after occurring in a `Refiner` abstraction. I couldn't figure out a way to get rid of needing to use a subprocess unfortunately (I explored using a custom linker with Wizer).

I chose to change the name from `Optimizer` to `Generator` because the Wasm fed into the `Optimizer` cannot be executed without causing a trap so that hinted that the name may not be quite correct. I opted to move running `wasm-opt` into the `Refiner` because that will provide me with useful flexibility to make additional changes between running Wizer and `wasm-opt` while in the parent process where I have access to the JS source code. If we were able to figure out a way where we don't need to create a subprocess, we would add the logic in the `Refiner` into the `Generator`. One of the `Generator` or `Refiner` will gain an additional Walrus pass to add new exported functions for #405.

Given I had to rewrite `create_statically_linked_module` anyway, I opted to make additional changes to try to make the control flow easier to follow. I also opted to pass the existing CLI args through when creating the subprocess to avoid needing to make changes there in the future if the CLI args were to change for whatever reason. I also opted to move the file I/O up to `main.rs` because we'd can centralize the file I/O to write the generated Wasm in `main.rs`.

The second big change is I moved the static `Generator`, the dynamic `Generator`, and the transforms into a `wasm_generator` module. I think that makes it more obvious which files are generating the Wasm and which file is generating the static code, the dynamic code, and the shared logic between the two.

And the third change I made was to make the dynamic module generation use a `Generator` struct similar to the static code. All of the code in the `generate` function is the same as what's in was in `module_generator` but I guess Git got confused. Changes for #405 will also add Walrus logic to add the new exported functions.

I could also split this PR up into three so let me know if you'd prefer that. But given it's only ~125 lines changed, that seemed like overkill.


saulecabrera (commented):
The direction looks good. I have a couple of general comments:

* Do we need to introduce structs for `Generator` and `Refiner`? I left a comment about this for the `Generator` but taking a look at the big picture, I don't think that the structs are needed, since they only borrow `JS` and then perform an action on top of it. Can these modules be replaced by functions instead?

* Can we remove the optimization pass from the `Refiner`? Currently the `refine` function does the optimization _and_ the transformation, but technically it doesn't need to happen that way, so I was thinking if we could move the optimization to its own function. For context, the reason why a struct was originally introduced is because the intention was to provide a single module that could be constructed through the builder pattern to represent all the transformations that you could apply to a module, one of them being the optimization:

```rust
Optimizer::new().optimize(true).option_a().option_b().transform()
```

But that didn't pan out in the end. We are always going to be applying the optimization anyway as far as I know so I think that there's no point in keeping the original `optimize` bool flag. But even if did want to skip the optimization, we could just avoid running the `optimize` function.

jeffcharles:
Discussed offline.

> Do we need to introduce structs for Generator and Refiner?

Nope! I can make those functions.

> Can we remove the optimization pass from the Refiner?

I'll be keeping the optimization pass in the `Refiner` (or equivalently named function) but I'll be removing the boolean flag since it's always set to true.

saulecabrera (approved):

PR #412, opened 2023-06-27
Add support for multiple exported functions

### LINKED ISSUE #405: Add support for exposing exported JS functions as Wasm exports

jeffcharles:
Fixes #405.

The major features of the PR are:
- `JS` gets a new `exports` method that uses `swc_core` to parse the source code and identify both exported functions and named exports that are functions
- `javy-core`'s `main.rs` gets `canonical_abi_realloc` and `javy.invoke` functions (both are stripped out of the final module). The `javy.invoke` function uses `JSContextRef`'s `eval_module` to import and invoke a provided JS function name.
- The static Wasm generator's `refine` function adds a new export to the Wasm module for each exported JS function. This export loads the name of the JS function from a passive data segment into an allocated memory address and then calls `javy.invoke`.
- `javy-core`'s `lib.rs` gets an `invoke` function which both evaluates QuickJS bytecode and then imports and invokes a JS function. This is because the module is not available in the JS runtime in the provider module so it needs to be evaluated first.
- The dynamic Wasm generator's `generate` function adds a new export to the Wasm module for each exported JS function. This export loads the bytecode into an allocated memory address and then loads the JS function name into a different allocated memory address and then calls `invoke` with both addresses.
- A new optional CLI flag `--wit` is introduced to specify a WIT file describing the exports for the JS module
- A new optional CLI flag `-n` is introduced to specify the name of the WIT world that describes the exports
- The WIT file provided by `--wit` is parsed and the `-n` flag is used to identify the world to retrieve export definitions from. An error is returned if the JS module does not contain the export.


crates/cli/src/js.rs
```diff
@@ -3,43 +3,61 @@
...
+        parser::{self, EsConfig, Syntax},
+    },
 };
 
 use crate::bytecode;
 
+#[derive(Clone, Debug)]
 pub struct JS {
-    source_code: Vec<u8>,
+    source_code: Rc<String>,
```
jeffcharles, line 31: I changed this to store an `Rc<String>` because SWC expects either a `String` or an `Rc<String>` as input to its parser. I debating performing the conversion in the `exports` method but the nice property of it being stored as an `Rc<String>` is cloning is very cheap and it's cheap to get `&[u8]` or `&str` representations from this format so I figured we may as well do the conversion ahead of time.

crates/cli/src/js.rs
```diff
@@ -48,4 +66,232 @@ impl JS {
...
+                        (f.function.params.is_empty(), f.function.is_generator),
+                    );
+                }
+                _ => continue,
+            }
+        }
+
+        let mut named_exported_functions = named_exports
+            .into_iter()
+            .filter_map(|(orig, exported)| {
```
jeffcharles, line 120: This is a little complicated because we have 3 separate cases for named exports. They can be:
- not a function
- a function we don't support (has arguments or is a generator)
- a function we support

And we need to drop the first case, error on the second, and figure out the exported name for the third case.

crates/cli/src/wasm_generator/static.rs
```diff
@@ -36,10 +39,94 @@ pub fn refine(mut wasm: Vec<u8>, js: &JS) -> Result<Vec<u8>> {
...
+
+struct Exports {
+    realloc_export: ExportId,
+    realloc_fn: FunctionId,
+    invoke_export: ExportId,
+    invoke_fn: FunctionId,
+    memory: MemoryId,
+}
+
+fn get_useful_exports(module: &walrus::Module) -> Exports {
```
jeffcharles, line 104: I feel like there's got to a better pattern to do what I'm trying to do in here. Namely, given a collection of things, find the ids for specific things that _should_ be in the collection.

crates/cli/tests/integration_test.rs
```diff
@@ -84,7 +84,7 @@ fn test_readme_script() {
 fn test_promises() {
     let mut runner = Runner::new("promise.js");
 
-    let (output, _, fuel_consumed) = run(&mut runner, &[]);
+    let (output, _, _) = run(&mut runner, &[]);
```
jeffcharles, line 87: This line is kind of unrelated to this PR but I got a warning when running the tests with the `experimental_event_loop` feature enabled so figured I'd fix the warning while I was in here.

crates/core/src/main.rs
```diff
@@ -29,5 +34,44 @@ pub extern "C" fn init() {
...
+///
+/// # Safety
+///
+/// * `fn_name_ptr` must reference a UTF-8 string with `fn_name_size` byte
+///   length.
+#[export_name = "javy.invoke"]
+pub unsafe extern "C" fn invoke(fn_name_ptr: *mut u8, fn_name_size: usize) {
+    let js_fn_name = str::from_utf8_unchecked(slice::from_raw_parts(fn_name_ptr, fn_name_size));
+    let runtime = unsafe { RUNTIME.take().unwrap() };
+    execution::invoke_function(&runtime, FUNCTION_MODULE_NAME, js_fn_name);
```
jeffcharles, line 76: Interestingly enough, if I run `execution::run_bytecode(runtime, bytecode);` in here before `invoke_function`, I see the top-level logic getting executed twice in the tests. However, it works as expected in `lib.rs` when `run_bytecode` is executed followed by `invoke_function`.

saulecabrera (commented):
Great work! Left some comments that I think should be addressed before landing this.

README.md
```diff
@@ -137,6 +137,28 @@ $ echo '{ "n": 2, "bar": "baz" }' | wasmtime index.wasm
 {"foo":3,"newBar":"baz!"}%   
 ```
 
+### Exporting functions
+
+For each exported JavaScript function, Javy will add an additional function export to the WebAssembly module. Exported functions with arguments and generators are not supported. Return values will also be dropped and not returned.
+
```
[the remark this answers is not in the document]
jeffcharles: True. We should probably also update the docs elsewhere to say that CommonJS modules are not supported given that we seem to get a lot of questions about why `require` isn't working.

crates/cli/src/wasm_generator/static.rs
jeffcharles: > I thought that Walrus would provide an api to achieve this easier, it seems that something like ModuleExports::by_name is what we need, similar to ModuleFunctions::by_name.

Yeah that would've been nice.

I did consider using `ModuleFunctions::by_name` but it's worth pointing out that the semantics with that method is different since it uses the `name` custom section and multiple functions can have the same name in that section and we don't control the name used.

What I could do is try to build a `Map<&str, &Export>` which would roughly correspond to how a `by_name` method would work, but I'd still need a pass to pull the underlying IDs out given the `ExportItem` variant on the export. I can experiment a little.

> can we rename this function to something else?

Yeah I was struggling with the name. I didn't really like `get_exports` because it's a subset of the exports. I didn't like `get_realloc_and_invoke_and_memory_ids`, especially given it would have to change depending on what's being returned. I'll experiment a little more but worst case, I'll leave a comment.

jeffcharles: I completely changed how I was going about this. I built a `HashMap<&str, &Export>` and was able to use the relatively new `let-else` syntax to get the ids I needed out. There's probably still some room for improvement but it meant I could get rid of the helper function and struct entirely.

README.md
```diff
@@ -137,6 +137,28 @@ $ echo '{ "n": 2, "bar": "baz" }' | wasmtime index.wasm
...
+```
+
+In the terminal:
+```bash
+$ javy compile index.js -o index.wasm
+$ wasmtime run --invoke foo index.wasm
+Hello world!
+Hello from foo!
+```
+
```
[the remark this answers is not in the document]
jeffcharles: I made a change to gate behind a CLI flag as suggested.

The only concern I have is I'm not sure how we would be able to transition this out of an experimental state. But I suppose we don't need to worry about that now.

jeffcharles:
Switched to WIT to identify exports so we can validate they're not intending to return any values.

saulecabrera (approved):
LGTM, should the version be updated to the next alpha?

jeffcharles:
> LGTM, should the version be updated to the next alpha?

I'll create a follow-up PR to update the minor version after this is merged.

PR #414, opened 2023-06-29
Javy-cli set exit code when javy fails

jeffcharles:
The `javy-cli` NPM package should exit with Javy's exit code after running Javy. If Javy fails to start, it should exit with a non-zero exit code. Other tools may inspect the exit code of `javy-cli` to determine if running it failed and set their exit code appropriately.

I'm also going to include a stderr line about checking if glibc is installed if a `ENOENT` error code is returned. A few things may cause that like if a binary for the wrong architecture is downloaded. But in our case, it's probably because glibc isn't available and there isn't any output if that's the case.

### DIFF
--- a/npm/javy-cli/index.js
+++ b/npm/javy-cli/index.js
@@ -20,13 +20,12 @@ async function main() {
 			await downloadBinary(version);
 		}
 	}
-	try {
-		childProcess.spawnSync(binaryPath(version), getArgs(), {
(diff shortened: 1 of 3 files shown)


jeffcharles:
I forgot doing this will break the NPM package CI. I'll fix that.

saulecabrera (approved):

surma (approved):

PR #419, opened 2023-07-06
Update javy-cli README and fix building from source

jeffcharles:
Updating the README so it's in-sync with what's in the code. I also noticed building from source wasn't working so I made a couple small fixes to get that working.

### DIFF
--- a/npm/javy-cli/index.js
+++ b/npm/javy-cli/index.js
@@ -15,7 +15,7 @@ async function main() {
 	const version = await getDesiredVersionNumber();
 	if (!(await isBinaryDownloaded(version))) {
 		if (process.env.FORCE_FROM_SOURCE) {
-			await buildBinary();
+			await buildBinary(version);
 		} else {
 			await downloadBinary(version);
 		}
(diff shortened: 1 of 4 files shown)


npm/javy-cli/index.js
```diff
@@ -51,7 +51,7 @@ async function downloadBinary(version) {
 	const targetPath = binaryPath(version);
 	const compressedStream = await new Promise(async (resolve) => {
 		const url = binaryUrl(version);
-		console.log(`Downloading ${NAME} ${version} to ${targetPath}...`);
+		console.log(`Downloading ${NAME} ${version} to ${targetPath}`);
```
jeffcharles, line 54: I removed the `...` because the name of the file ends with a something that looks like `v1.1.1` which is rendered as `v1.1.1...` which makes it harder to figure out where the path ends.

npm/javy-cli/index.js
```diff
@@ -150,10 +150,11 @@ function getArgs() {
...
 }
 
-async function buildBinary() {
+async function buildBinary(version) {
 	const repoDir = cacheDir("build", NAME);
 	try {
 		console.log(`Downloading ${NAME}'s source code...`);
+		fs.rmSync(repoDir, { recursive: true });
 		childProcess.execSync(
 			`git clone https://github.com/${REPO} ${repoDir}`
```
jeffcharles, line 159: This ignores the version argument which is probably incorrect. Arguably the right thing to do is add support in the code for running a plain `javy` executable instead of including a version number. I didn't want to spend more time doing that given I'm not sure how frequently this is used in practice, if at all. We can opt to fix if someone complains.

saulecabrera (approved):

surma (approved):

jeffcharles:
I just realized I forgot to update the package-lock. Will do that before merging.

PR #423, opened 2023-07-07
Better error when attempting to export default arrow function

### LINKED ISSUE #422: Exported default function results in unhelpful error message
Operating system: N/A
Processor architecture: N/A
Rust version: N/A
Javy version: N/A

## Problem

When using a WIT of:
```
package function:impl

world world-name {
  export named: func()
  export default: func()
}
```

and a JS of:

```JS
export function named() {};
export default () => {};
```

Javy outputs:

```
Error: JS module does not export default
```

We should output a module with an exported function named `default`.

This matches behaviour like the following in Node:

foo.mjs:
```js
export default () => { ... }
```

bar.mjs
```js
import { default as baz } from './foo.mjs'
```

jeffcharles:
Fixes #422. The error message will now clearly indicate that exporting default arrow functions is not supported.

### DIFF
--- a/crates/cli/src/js.rs
+++ b/crates/cli/src/js.rs
@@ -105,6 +105,9 @@ impl JS {
                         }
                     }
                 }
+                ModuleItem::ModuleDecl(ModuleDecl::ExportDefaultExpr(e)) if e.expr.is_arrow() => {
+                    bail!("Exported default arrow functions are not supported");
+                }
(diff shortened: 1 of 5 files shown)


jeffcharles:
Going to close because I realized `default` is always an invalid function identifier in JS and JS supports a syntax that looks like `import { default as foo } from './bar.mjs'` so we probably should support a default export of an arrow function with a Wasm function name of `default`.

PR #424, opened 2023-07-07
Fix multiword exports

### LINKED ISSUE #420: Support function exports with multiple words
## What is the idea?

Support export names that have multiple words.

For example,
```
world example {
  export foo-bar: func()
}
```

```js
export function fooBar() {
  ...
}
```

The code above would result in a Wasm function export named `foo-bar` to align with what's in the WIT.

At the moment, WIT only allows for kebab-casing when naming exports and Javy requires an exact match between the name in the WIT file and the name in the source code. QuickJS does not support exporting function names with kebab-casing. These two restrictions make it impossible to export function names that have multiple words.

jeffcharles:
Fixes #420.

We translate the kebab-cased identifiers in the WIT file to camel-cased identifiers in the JavaScript and retain the kebab-cased identifiers for the name of the Wasm function exports since the WIT describes the export names for the Wasm.

### DIFF
--- a/crates/cli/src/exports.rs
+++ b/crates/cli/src/exports.rs
@@ -9,7 +9,7 @@ pub struct Export {
     pub js: String,
 }
 
-pub fn determine_js_exports(js: &JS, wit: &Path, wit_world: &str) -> Result<Vec<Export>> {
+pub fn process_exports(js: &JS, wit: &Path, wit_world: &str) -> Result<Vec<Export>> {
     let js_exports = js.exports()?;
     wit::parse_exports(wit, wit_world)?
         .into_iter()
--- a/crates/cli/src/exports.rs
+++ b/crates/cli/src/exports.rs
@@ -1,0 +1,28 @@
+use anyhow::{anyhow, Result};
+use convert_case::{Case, Casing};
+use std::path::Path;
+
+use crate::{js::JS, wit};
+
+pub struct Export {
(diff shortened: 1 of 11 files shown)


crates/cli/src/exports.rs
```diff
@@ -0,0 +1,28 @@
...
+use std::path::Path;
+
+use crate::{js::JS, wit};
+
+pub struct Export {
+    pub wit: String,
+    pub js: String,
+}
+
+pub fn determine_js_exports(js: &JS, wit: &Path, wit_world: &str) -> Result<Vec<Export>> {
```
saulecabrera, line 12: Can we add some documentation here or rename this function to something that indicates that we are _renaming_ and _searching_ the relevant exports (e.g `process_exports`)? At a first glance it was not evident, at least to me, that here's where all the casing is happening.

saulecabrera (approved):

crates/cli/src/exports.rs
jeffcharles: My thinking with `determine_js_exports` was it was getting a list of WIT exports and _determining_ which JS export corresponded to each of them. There isn't any renaming going on, the data structure that gets sent back is essentially `(wit_export, js_export)`. That said, I'll use a different name given the current one seems to not be clear. `process_exports` seems fine to me so I'll update to use that.

PR #429, opened 2023-07-11
Remove make target for downloading WASI SDK

jeffcharles:
We no longer need to download the WASI SDK in CI out-of-band since we download it in https://github.com/bytecodealliance/javy/blob/main/crates/quickjs-wasm-sys/build.rs. This doesn't remove the existing support for using a custom WASI SDK. This also helps clean up the CI workflow a little.


saulecabrera (commented):
In https://github.com/bytecodealliance/javy/pull/383 we discussed (offline) that downloading via `build.rs` would not necessarily be the canonical way of using a given wasi-sdk version. So I think I'm in favor of removing downloading wasi-sdk in CI to reduce complexity, but I'm not sure about removing the make target, since the `quickjs-wasm-sys` build process can still be set to use a custom path via WASI_SDK_PATH, which could be useful to quickly test other wasi-sdk releases.

jeffcharles:
> I'm not sure about removing the make target, since the quickjs-wasm-sys build process can still be set to use a custom path via WASI_SDK_PATH, which could be useful to quickly test other wasi-sdk releases

I don't think I'm understanding. The build process for `quickjs-wasm-sys` still uses `WASI_SDK_PATH` to find a custom installed SDK, so no change there. The make target does **not** download the SDK to `WASI_SDK_PATH` if `WASI_SDK_PATH` is set, instead it checks if there's something in `WASI_SDK_PATH` and exits with an exit code of 0 if there is and an exit code of 1 if there isn't. The quickjs-wasm-sys build file performs an identical check.

Specifically the make target follows this logic if `WASI_SDK_PATH` is set: https://github.com/bytecodealliance/javy/blob/9502f790a6646ea2d482e06feee61a335549104b/install-wasi-sdk.sh#L12-L19

And the build file does this: https://github.com/bytecodealliance/javy/blob/9502f790a6646ea2d482e06feee61a335549104b/crates/quickjs-wasm-sys/build.rs#L124-L139

So the make target and build file have the same logic if `WASI_SDK_PATH` is set. And with this change, it will still be easy to test with a different version of the WASI SDK by either setting the `WASI_SDK_PATH` env var or changing the value in the quickjs-wasm-sys build file.

jeffcharles:
Anyway, I'll split off the Makefile piece from this.

saulecabrera:
> I don't think I'm understanding. The build process for quickjs-wasm-sys still uses WASI_SDK_PATH to find a custom installed SDK, so no change there.

Yeah, I'm well aware that there's no change there and that's primarily the reason for my comment. What I really want is to be able to do:

```sh
make dowload-wasi-sdk && QUICKJS_WASM_SYS_WASI_SDK_PATH=... cargo build ...
``` 

Instead of having to change the `build.rs` and having to recompile. So even though the script doesn't download wasi-sdk to `WASI_SDK_PATH` it doesn't void the fact that it's still useful to have that script around.

saulecabrera (approved):

Issue #324, opened 2023-04-25
Audit CI to find opportunities to have it finish faster


jianghong:
There may be opportunities to have Javy's CI execute faster on PRs. Doing this means we get faster feedback.

jeffcharles:
- Using Linux instead of MacOS for CI resulted in a ~2x speedup according to https://github.com/jeffcharles/javy/actions/runs/5401996292 (28 minutes on that branch vs 54 minutes in main)

ubiquitous-dev:
Looks like the tests are taking a majority of the CI time.  Is there a way to parallelize wasi tests? or is that already happening, and the overhead of running the tests is just large?  If that's the case, one option would be to shard the tests and have multiple jobs run in parallel using a matrix.

jeffcharles:
> Looks like the tests are taking a majority of the CI time

The step labelled _Tests_ in CI performs a number of distinct operations including compilation. Here's a breakdown of  the timings I'm seeing from [the latest CI run](https://github.com/bytecodealliance/javy/actions/runs/5508808121/jobs/10040599862):

| Step | Time |
| --- | --- |
| Checkout | 46s |
| Restore cargo cache | 30s |
| Restore target cache | 58s |
| Install wasmtime-cli | 1s |
| Install cargo-wasi | 3s |
| Install cargo-hack | 1s |
| Install wizer | 0s |
| Install wasi-sdk | 3s |
| Compile quickjs-wasm-rs | 56s |
| Test quickjs-wasm-rs | 5s |
| Compile javy | 40s |
| `cargo hack wasi test --package=javy-apis --each-feature -- --nocapture` | 16s |
| Compile javy-core for testing | 1s |
| Compile release of javy-core | 56s |
| Wizer javy-core | 1s |
| Compile javy-cli | 17m 8s |
| Test javy-cli | 1m 55s |
| Re-compile javy-cli for WPT (enables promises feature) | 6m 39s |
| WPT | 14s |
| Check benchmarks | 2s |
| Lint | 1m 25s |
| Post target cache | 0s |
| Post cargo cache | 0s |

It looks to me like the compilation of `javy-cli` (and the recompilation for WPT) is what's using the majority of the CI time. I think it makes sense to look focus on ways to reduce that time or parallelize the recompilation for WPT.

[...]

PR #438, opened 2023-07-17
Disable LTO for check-benchmarks and CLI clippy

jeffcharles:
Part of #324. Now that we compile the CLI for tests with LTO disabled, this reduces CI time by 13 minutes according to [a build with LTO enabled taking 14m 55s](https://github.com/jeffcharles/javy/actions/runs/5546416904/jobs/10128772493) and [a build without LTO taking 1m 15s](https://github.com/jeffcharles/javy/actions/runs/5547801410/jobs/10130430402). I'm also going to disable LTO for clippy since that also requires compilation.


surma:
WTF

surma (approved):

PR #443, opened 2023-07-21
Remove make from GH workflows

jeffcharles:
Stop using `make` in GitHub workflows. This could make it easier to switch to using MacOS to build javy-core and Ubuntu for CLI builds and testing in CI to speed up CI. This would be because we would no longer be depending on the Makefile to handle the dependency resolution for running the javy-core crate before building or testing javy-cli and we're instead making that explicit in the CI workflow. The reason for continuing to build javy-core in MacOS is we see different fuel usage for the same JS but using javy-core built on different operating systems and MacOS has the lowest fuel usage. But I've measured that Ubuntu seems to compile the CLI faster. I have branch where I tried using Ubuntu for CLI builds and testing and MacOS for `javy-core` but the changes look messy when using `make` for some things and not for others.

Also we should probably move away from using `make` where possible anyway since it's not idiomatic in Rust projects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,11 +54,16 @@ jobs:
         uses: taiki-e/install-action@cargo-hack
 
       - name: Tests
-        # Everything but WPT
-        run: make test-quickjs-wasm-rs test-javy test-apis test-core test-cli
+        run: |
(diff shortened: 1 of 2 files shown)


surma (approved):
Always happy to see less Makefile :D

PR #444, opened 2023-07-24
Stop using subprocess when generating Wasm

jeffcharles:
I was looking at something earlier today and got inspired. We can use the `make_linker` method on Wizer to pass a custom WASI context with the stdin set to the JS source code. This means we don't need to create a subprocess to set the stdin to the JS source code for Wizer. I think not using a subprocess reduces the complexity involved with generating a statically linked Wasm module. It also opens the door for potentially switching to using a deterministic WASI context when creating JS modules in the future.

The one additional piece of complexity that gets pulled in is we now have to use a mutable static `OnceLock` in the static generator code which means we would not be able to run that code concurrently. I don't know when or why we would want to do that so this seems like an acceptable tradeoff to me.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -5,16 +5,13 @@ mod js;
 mod wasm_generator;
 mod wit;
 
-use crate::commands::{Command, CompileCommandOpts, EmitProviderCommandOpts};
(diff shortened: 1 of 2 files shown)


surma (approved):
Nice one!

PR #452, opened 2023-08-15
Test javy-cli on Node versions 16, 18, and 20

jeffcharles:
We want to ensure `javy-cli` works as expected against the last few stable Node versions.


jeffcharles:
Will update required checks if this passes review.

saulecabrera (approved):

PR #455, opened 2023-08-16
Version crates to 2.0.0 for release

jeffcharles:
Bumping `quickjs-wasm-rs`, `javy`, and `javy-apis` to 2.0.0 for release.

I realized going through the CHANGELOG that we missed a breaking change in `javy` because it re-exports `quickjs-wasm-rs` so `javy` needs to go to 2.0.0 as well. And we also had an informal policy that was never documented to keep `javy-apis` in sync with `javy` so that also has to go to 2.0.0 as well. I've included an update to the versioning policy in this PR.

I'm also opting to remove the version diffs in the changelog because they're very difficult to properly maintain. We don't use tags for crate releases, we can't know the commit hash on the `main` branch before merging, and branches may not reflect the contents of a crate when merged. I'm open to revisiting if there's a way to do this.

We might need smarter tools or better documentation to consult for ensuring versions are set correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1343,7 +1343,7 @@ dependencies = [
 
 [[package]]
 name = "javy"
-version = "1.0.0"
+version = "2.0.0"
 dependencies = [
  "anyhow",
  "quickjs-wasm-rs",
@@ -1354,7 +1354,7 @@ dependencies = [
 
 [[package]]
 name = "javy-apis"
-version = "1.1.0-alpha.1"
+version = "2.0.0"
 dependencies = [
  "anyhow",
  "fastrand",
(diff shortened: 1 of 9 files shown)


saulecabrera (approved):
> We might need smarter tools or better documentation to consult for ensuring versions are set correctly.

Or another option is to not follow semver. Trying to get semver right seems to introduce a considerable amount of overhead on maintenance (or tooling), which we don't have and I don't think we'll invest on in the near future, which makes me think that the cost outweighs the benefits. It might be simpler to treat each release as a new "major" version altogether. 

That said, I'm not expecting that we decide a new versioning scheme as part of this PR, but I think that if we have at least one more instance in which versioning is proven to be difficult to get right, we might want to seriously consider it.

PR #461, opened 2023-08-18
Move allocation functions into javy lib

jeffcharles:
This does two things:
1. Moves the realloc and free implementations from the core crate into the `javy` crate under the publicly exported `alloc` module
2. Adds a cargo feature named `export_alloc_fns` which, when enabled, will export the realloc and free functions from the produced Wasm module

This will allow users of the Javy crate to share memory allocation and free functions instead of potentially having to write their own. I opted to put the function exports behind a cargo feature because someone using the Javy crate may want to expose their own allocation and free functions instead with their allocation logic and wouldn't want someone calling our functions or if they want to export them under a different name.

### DIFF
--- a/crates/cli/src/wasm_generator/static.rs
+++ b/crates/cli/src/wasm_generator/static.rs
@@ -45,19 +45,21 @@ pub fn generate(js: &JS, exports: Vec<Export>) -> Result<Vec<u8>> {
 
     let mut module = transform::module_config().parse(&wasm)?;
 
-    let (realloc, invoke, memory) = {
+    let (realloc, free, invoke, memory) = {
         let mut exports = HashMap::new();
         for export in module.exports.iter() {
             exports.insert(export.name.as_str(), export);
         }
         (
             *exports.get("canonical_abi_realloc").unwrap(),
+            *exports.get("canonical_abi_free").unwrap(),
             *exports.get("javy.invoke").unwrap(),
             *exports.get("memory").unwrap(),
         )
     };
 
     let realloc_export = realloc.id();
+    let free_export = free.id();
     let invoke_export = invoke.id();
 
     if !exports.is_empty() {
@@ -69,6 +71,7 @@ pub fn generate(js: &JS, exports: Vec<Export>) -> Result<Vec<u8>> {
 
     // We no longer need these exports so remove them.
     module.exports.delete(realloc_export);
+    module.exports.delete(free_export);
     module.exports.delete(invoke_export);
 
     let wasm = module.emit_wasm();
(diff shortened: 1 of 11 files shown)


crates/cli/src/wasm_generator/static.rs
```diff
@@ -69,6 +71,7 @@ pub fn generate(js: &JS, exports: Vec<Export>) -> Result<Vec<u8>> {
 
     // We no longer need these exports so remove them.
     module.exports.delete(realloc_export);
+    module.exports.delete(free_export);
```
jeffcharles, line 74: I'm doing this for consistency with us hiding the realloc function

crates/javy/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub use config::Config;
 pub use quickjs_wasm_rs as quickjs;
 pub use runtime::Runtime;
 
+pub mod alloc;
```
saulecabrera, line 43: Why don't we instead add `[cfg(feature = "export_alloc_fns")]` here and remove the cfg directive from each of the function definitions?

crates/javy/src/alloc.rs
```diff
@@ -7,6 +7,21 @@ use std::ptr::copy_nonoverlapping;
 // non-zero to indicate success.
 const ZERO_SIZE_ALLOCATION_PTR: *mut u8 = 1 as _;
 
+/// 1. Allocate memory of new_size with alignment.
+/// 2. If original_ptr != 0
+///   a. copy min(new_size, original_size) bytes from original_ptr to new memory
+///   b. de-allocate original_ptr
+/// 3. return new memory ptr
```
saulecabrera, line 14:
```suggestion
/// 3. Return new memory ptr.
```

saulecabrera: (I'm aware that this is not your change, but I thought why not improve it, given that we're here)

saulecabrera (commented):
Generally looks good to me. There are a couple of minor fixes that we could add before landing it.

crates/javy/src/lib.rs
jeffcharles: One small reason is if they don't enable the `export_alloc_fns` feature, they can still add exported functions in their own crate to call `canonical_abi_realloc` and `canonical_abi_free` but export them with a name of their own choosing instead of with the export name we chose.

saulecabrera (approved):
Discussed the cfg feature offline. I think that a comment will help with some of the confusion here, it feels a bit odd that Rust/Cargo requires a feature/export_name in order to keep those symbols.

Issue #486, opened 2023-10-09
Get wrapped rust value in quickjs-wasm-rs 2.0.1


ullgren:
## What is your question?
First let me confess that I'm a beginner to rust, so I appoligize in advance if this is a newbie question.
 
In version 0.1.4 of the  quickjs-wasm-rs crate it was possible to create a `Value` by wrapping a rust value and then adding callbacks on the property. In this call back one could then use `this.get_rust_value()` to get the wrapped rust value. See example below. 

I'm struggling to migrate this example into newer versions of the crate, such as 2.0.1. It seems that while the `JSContextRef::wrap_rust_value()` function is still present there is no obvious way to get the rust value back from the resulting `JSValueRef`. I've tries using various versions of `try_into` and similar. 

Example that compiles and run as expected with quickjs-wasm-rs 0.1.4

```rust
use quickjs_wasm_rs::Context;

struct SomeData {
    counter: i32,
    message: String,
}

fn main() {
    let context = Context::default();
    let some_data = SomeData {
        counter: 42,
        message: "Hello World!".to_string(),
    };
    let some_data_js_value = context.wrap_rust_value(some_data).unwrap();

    some_data_js_value
        .set_property(
            "increment",
            context
                .wrap_callback(move |ctx, this, _args| {
                    let mut some_data = this.get_rust_value::<SomeData>().unwrap().borrow_mut();
                    some_data.counter += 1;
                    ctx.undefined_value()
                })
                .unwrap(),
        )
        .unwrap();
    some_data_js_value
        .set_property(
            "print",
            context
                .wrap_callback(move |ctx, this, _args| {
                    let some_data = this.get_rust_value::<SomeData>().unwrap().borrow_mut();
                    println!(
                        "Some data: count {} message {}",
                        some_data.counter, some_data.message
                    );
                    ctx.undefined_value()
                })
                .unwrap(),
        )
        .unwrap();
    let global = context.global_object().unwrap();
    global.set_property("somedata", some_data_js_value).unwrap();

    context
        .eval_global("index.js", "somedata.increment(); somedata.print('hello');")
        .unwrap();
}
```

saulecabrera:
[...]

PR #490, opened 2023-10-19
Make `JSContextRef::wrap_rust_value` private

saulecabrera:
## Description of the change
While investigating https://github.com/bytecodealliance/javy/issues/486 I realized that this function is still public, which shouldn't be the case as it's only used for wrapping closure types, which is an internal detail about the crate.

## Checklist

### DIFF
--- a/crates/quickjs-wasm-rs/CHANGELOG.md
+++ b/crates/quickjs-wasm-rs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- Make `JSContextRef::wrap_rust_value` private. Similar to
+  `context::get_rust_value` this function is simply an internal detail.
+
 ## [2.0.1] - 2023-09-11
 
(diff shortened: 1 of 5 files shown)


crates/quickjs-wasm-rs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- Make `JSContextRef::wrap_rust_value` private. Similar to
```
jeffcharles, line 11: Should this be considered removed? The function is still there, but from the library consumer's point of view, it's a feature that's been removed from the public API. I don't feel strongly about it though.

jeffcharles (approved):

crates/quickjs-wasm-rs/CHANGELOG.md
saulecabrera: I consider it, but for someone reading the code, or even comparing differences between both versions, I concluded that this is probably the best wording to capture what happened.

PR #584, opened 2024-01-05
Fixed no-source-compression to compress at level 0 instead of no-compression

orisano:
rel: #581
## Description of the change
Fixed no-source-compression to compress at level 0 instead of no-compression

https://github.com/bytecodealliance/javy/pull/581#issuecomment-1878925292

## Checklist

### DIFF
--- a/crates/cli/src/js.rs
+++ b/crates/cli/src/js.rs
@@ -54,13 +54,13 @@ impl JS {
         bytecode::compile_source(self.source_code.as_bytes())
     }
 
-    pub fn compress(&self) -> Result<Vec<u8>> {
+    pub fn compress(&self, quality: i32) -> Result<Vec<u8>> {
         let mut compressed_source_code: Vec<u8> = vec![];
         enc::BrotliCompress(
             &mut Cursor::new(&self.source_code.as_bytes()),
             &mut compressed_source_code,
             &BrotliEncoderParams {
-                quality: 11,
+                quality,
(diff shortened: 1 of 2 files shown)


jeffcharles:
I'm not convinced not being able to tell whether the contents of the section are brotli compressed or not will be a problem in practice. You're correct in that there isn't a 100% correct way to tell whether an array of bytes is JS source code or a Brotli compressed byte array. But the source code custom section that Javy creates is purely for informational purposes and can't and should not be relied on. Someone can take a Javy produced Wasm module and tamper with that section or remove it entirely.

My concern with performing a `0` level Brotli compression with the `--no-source-compression` flag is that the source code is still compressed, just with a lower Brotli quality, so the flag name would be misleading. Hypothetically, we could change the flag to something like `--fast-source-compression`. That might be okay. For context, I also don't want to expose that we're using Brotli in the command line flags either because it makes it harder to change the compression algorithm later. The other tradeoff with using a Brotli compression level of `0` as opposed to no compression is it's still slower at compile time than not doing any compression. From a quick check it looks like pessimistically, Brotli will compress on the order of 3.4 megabytes per second[^1]. Sounds like that's fine for your use case though.

[^1]: https://blog.cloudflare.com/results-experimenting-brotli#compressionspeed

orisano:
I understand.
If it does not matter if the source code is compressed or not distinguishable, I will close this PR!
Thanks!

Issue #605, opened 2024-02-10
Expose Rust structs to JavaScript as classes


rahul007-bit:
I found a [line](https://github.com/awslabs/llrt/blob/main/src/http/response.rs#L67) in AWS LLRT and was wondering if we can do the same in Javy too, create a struct that JS can access as a class.

saulecabrera:
Hi @rahul007-bit, even though a pattern like this doesn't exist today in Javy, in theory we could explose a `class` macro (or similar) that allows you to achieve a similar behavior. Do you have a specific use-case in mind?

rahul007-bit:
I am working on a project that utilizes [Jsonnet](https://www.npmjs.com/package/@hanazuki/node-jsonnet), but unfortunately, it is not compatible with QuickJS. Therefore, I am attempting to integrate it as a built-in module within the runtime environment. At present, I am implementing this integration using the `eval_module` and `set_property` functions.

Here is the Rust code snippet I am currently employing:

```rust
static JSONNET: &str = include_str!("./index.js");

pub(super) struct Jsonnet;

impl JSApiSet for Jsonnet {
  fn register(&self, runtime: &javy::Runtime, _config: &crate::APIConfig) -> anyhow::Result<()> {
    let context = runtime.context();
    let global = context.global_object()?;
    global.set_property(
       "__jsonnet_evaluate_snippet",
       context.wrap_callback(jsonnet_evaluate_snippet_callback())?,
    )?;

    if let Err(err) = context.eval_module("jsonnet", JSONNET) {
      eprintln!("Error loading the path shim: {:?}", err);
    }
    Ok(())
  }
}
```

And here is the JavaScript class definition that corresponds to the Rust implementation:

```js
const __jsonnet_evaluate_snippet = globalThis.__jsonnet_evaluate_snippet;

class Jsonnet {
  evaluateSnippet(snippet) {
    return __jsonnet_evaluate_snippet(snippet);
  }
}
Reflect.deleteProperty(globalThis, "__jsonnet_evaluate_snippet");

export default Jsonnet;
```

However, when I attempt to import this module in JavaScript, the application encounters an error and crashes.

```js
import Jsonnet from "jsonnet";

addEventListener("fetch", (event) => {
  event.respondWith(handleRequest(event.request));
});

async function handleRequest(request) {
  return new Response("Hello world", { status:  200 });
}
```

[...]

PR #610, opened 2024-02-27
Add escape hatches to and from `quickjs_wasm_sys`

surma:
While `quickjs_wasm_rs` provides a much more ergonomic API, there are some features of QuickJS that are not exposed. For example, exotic object behaviors or QuickJS’s execution timeout.

In the long-run, it would be great to have those in `quickjs_wasm_rs`, but in the meantime, it seems like a good idea regardless to me expose escape hatches to the low-level crate.

Specifically: 
- Expose the `wasm_quick_sys` module from `quickjs_wasm_rs` to avoid users having to add that crate themselves and potentially ending up with different version.
- Add a function to convert raw QuickJS value (`u64`) into a `JSValueRef`
- Add functions to convert a `*mut JSContext` to a `JSContextRef` and vice versa (in fact, `*mut JSContext` was already exposed via [`new_callback`](https://docs.rs/quickjs-wasm-rs/latest/quickjs_wasm_rs/struct.JSContextRef.html#method.new_callback), but hard to utilize without the other parts).

### DIFF
--- a/crates/quickjs-wasm-rs/src/js_binding/context.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/context.rs
@@ -146,6 +146,9 @@ impl JSContextRef {
     /// # Safety
     /// While calling this function is safe, using it’s return value has to be
     /// done with a sufficitenly deep understanding of QuickJS and [quickjs_wasm_sys].
+    ///
+    /// This function is not part of the crate’s semver API contract.
+    #[cfg(feature = "export-sys")]
     pub unsafe fn as_raw(&self) -> *mut JSContext {
         self.inner
     }
@@ -155,6 +158,9 @@ impl JSContextRef {
     /// # Safety
     /// The caller has to ensure that the returned `JSContextRef` is the only one
     /// used throughout its lifetime.
+    ///
+    /// This function is not part of the crate’s semver API contract.
+    #[cfg(feature = "export-sys")]
     pub unsafe fn from_raw(inner: *mut JSContext) -> JSContextRef {
         JSContextRef { inner }
     }
--- a/crates/quickjs-wasm-rs/src/js_binding/context.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/context.rs
@@ -141,6 +141,15 @@ impl JSContextRef {
         JSValueRef::new(self, raw)
     }
 
+    /// Returns the raw pointer to the underlying [quickjs_wasm_sys::JSContext].
+    pub unsafe fn as_raw(&self) -> *mut JSContext {
+        self.inner
+    }
+
+    pub unsafe fn from_raw(inner: *mut JSContext) -> JSContextRef {
+        JSContextRef { inner }
+    }
+
     /// Compiles JavaScript to QuickJS bytecode with an ECMAScript module scope.
     ///
     /// # Arguments
--- a/crates/quickjs-wasm-rs/src/js_binding/context.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/context.rs
@@ -142,10 +142,19 @@ impl JSContextRef {
     }
 
     /// Returns the raw pointer to the underlying [quickjs_wasm_sys::JSContext].
+    ///
+    /// # Safety
+    /// While calling this function is safe, using it’s return value has to be
+    /// done with a sufficitenly deep understanding of QuickJS and [quickjs_wasm_sys].
     pub unsafe fn as_raw(&self) -> *mut JSContext {
         self.inner
     }
 
+    /// Creates a `JSContextRef` from a pointer to a [quickjs_wasm_sys::JSContext].
+    ///
+    /// # Safety
+    /// The caller has to ensure that the returned `JSContextRef` is the only one
+    /// used throughout its lifetime.
     pub unsafe fn from_raw(inner: *mut JSContext) -> JSContextRef {
         JSContextRef { inner }
     }
--- a/crates/quickjs-wasm-rs/src/js_binding/value.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/value.rs
@@ -58,6 +58,10 @@ impl<'a> JSValueRef<'a> {
         Self { context, value }
     }
 
+    /// Creates a new `JSValueRef` from a `u64`, which is QuickJS’s internal representation.
+    ///
+    /// # Safety
+    /// The caller has to ensure that the given value is valid and belongs to the context.
     pub unsafe fn from_raw(context: &'a JSContextRef, value: JSValueRaw) -> Self {
         JSValueRef::new_unchecked(context, value)
     }
--- a/crates/quickjs-wasm-rs/src/js_binding/value.rs
+++ b/crates/quickjs-wasm-rs/src/js_binding/value.rs
@@ -58,6 +58,10 @@ impl<'a> JSValueRef<'a> {
         Self { context, value }
     }
 
+    pub unsafe fn from_raw(context: &'a JSContextRef, value: JSValueRaw) -> Self {
+        JSValueRef::new_unchecked(context, value)
+    }
+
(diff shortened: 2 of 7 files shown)


crates/quickjs-wasm-rs/src/js_binding/value.rs
```diff
@@ -187,8 +195,12 @@ impl<'a> JSValueRef<'a> {
 
     /// Converts the JavaScript value to a string if it is a string.
     pub fn as_str(&self) -> Result<&str> {
-        let buffer = self.as_wtf8_str_buffer();
-        str::from_utf8(buffer).map_err(Into::into)
+        if self.is_str() {
```
surma, line 198: This is a drive-by fix which I probably should have put in another PR. The function description says it only converts to a string if the JS value is a string, but the implementation never checked if it’s a string (as far as I can tell). This has led to some weird bugs for me where this function returned an `Ok` value, but didn’t contain anything.

saulecabrera (commented):
>  For example, exotic object behaviors or QuickJS’s execution timeout.

Could you add a bit more detail about which APIs you'd like to see exposed? I'm not very familiar with "exotic object behaviours" for example. 


Generally my stance is that `quickjs_wasm_rs` should be as safe as it can be; I think this change is not changing that fundamental principle given that the safety around the methods is correctly annotated and we're not enforcing the usage of unsafe code, it's pretty much opt-in. 

I'd be good though to have a list of features that you think we're missing in  `quickjs_wasm_rs` to be able to get get rid of these "escape hatches", so I'd appreciate if you could list those more concretely if you can.

crates/quickjs-wasm-rs/src/js_binding/context.rs
```diff
@@ -141,6 +141,24 @@ impl JSContextRef {
...
+
+    /// Creates a `JSContextRef` from a pointer to a [quickjs_wasm_sys::JSContext].
+    ///
+    /// # Safety
+    /// The caller has to ensure that the returned `JSContextRef` is the only one
+    /// used throughout its lifetime.
+    pub unsafe fn from_raw(inner: *mut JSContext) -> JSContextRef {
+        JSContextRef { inner }
+    }
+
```
saulecabrera, line 161: Could you add a comment to each, stating that given the safety implications of these methods, we're not making any commitments on maintaining them and that they can be removed in the future?

crates/quickjs-wasm-rs/src/js_binding/value.rs
```diff
@@ -58,6 +58,14 @@ impl<'a> JSValueRef<'a> {
         Self { context, value }
     }
 
+    /// Creates a new `JSValueRef` from a `u64`, which is QuickJS’s internal representation.
+    ///
+    /// # Safety
+    /// The caller has to ensure that the given value is valid and belongs to the context.
+    pub unsafe fn from_raw(context: &'a JSContextRef, value: JSValueRaw) -> Self {
```
saulecabrera, line 65: Similar here, could you add a comment stating that this method could be removed in the future?

crates/quickjs-wasm-rs/src/lib.rs
```diff
@@ -66,6 +66,8 @@ mod js_binding;
 mod js_value;
 mod serialize;
 
+pub use quickjs_wasm_sys;
```
saulecabrera, line 69: Could you add a comment regarding the safety of using this re-export?

surma: Is using this export inherently unsafe? All the functions by bindgen are already marked unsafe, right?

Or do you want the same comment here as above that the re-export might disappear in the future?

saulecabrera: Sorry yeah, not safety, but around maintenance and it might disappear in the future.

surma:
> I'd be good though to have a list of features that you think we're missing in quickjs_wasm_rs to be able to get get rid of these "escape hatches", so I'd appreciate if you could list those more concretely if you can.

I mean the short answer is: All QuickJS features 😅 

Specifically for my explorations right now, I want to make use of the following QuickJS features:
- [`JS_SetInterruptHandler`](https://docs.rs/quickjs-wasm-sys/latest/quickjs_wasm_sys/fn.JS_SetInterruptHandler.html) which calls a callback in regular intervals that can choose to terminate the execution (by returning `true`).
- [Exotic Methods](https://docs.rs/quickjs-wasm-sys/latest/quickjs_wasm_sys/struct.JSClassExoticMethods.html), which are a part of a QuickJS [class definition](https://docs.rs/quickjs-wasm-sys/latest/quickjs_wasm_sys/struct.JSClassDef.html). It effectively allows you to implement a JS class whose shape and implementation is backed by Rust. I’m currently working around this by using the [ES6 Proxy API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy), but that is a level of indirection I’d like to avoid in the future :)

In general I think these escape hatches are valuable and I don’t think they really pose a risk, but it’s totally your call.

saulecabrera:
> In general I think these escape hatches are valuable and I don’t think they really pose a risk, but it’s totally your call.

I think they are too IMO, given that we currently don't cover the entire surface area of QuickJS API. But in the future, ideally we should, hence my comment about not making any stability guarantees. And as mentioned above, they are not forcing anyone to use any unsafe operations. 

> [Exotic Methods](https://docs.rs/quickjs-wasm-sys/latest/quickjs_wasm_sys/struct.JSClassExoticMethods.html), which are a part of a QuickJS [class definition](https://docs.rs/quickjs-wasm-sys/latest/quickjs_wasm_sys/struct.JSClassDef.html). It effectively allows you to implement a JS class whose shape and implementation is backed by Rust

Oh ok, that makes sense. This is probably going to be useful for https://github.com/bytecodealliance/javy/issues/605

jeffcharles:
Would there be any value in putting the re-export of `quickjs-wasm-sys` and the escape hatches behind a cargo feature? That might help communicate the stability guarantees in code as well as in documentation.

surma:
@jeffcharles I don’t think I have seen that be done before, but I like the suggestion.

crates/quickjs-wasm-rs/src/lib.rs
surma: As it turns out, doc-strings are not really a thing for re-exports (at least not when I check `cargo doc` locally). I hope the feature gate addresses your concern?

crates/quickjs-wasm-rs/src/js_binding/value.rs
```diff
@@ -187,8 +206,12 @@ impl<'a> JSValueRef<'a> {
 
     /// Converts the JavaScript value to a string if it is a string.
     pub fn as_str(&self) -> Result<&str> {
-        let buffer = self.as_wtf8_str_buffer();
-        str::from_utf8(buffer).map_err(Into::into)
+        if self.is_str() {
+            let buffer = self.as_wtf8_str_buffer();
+            str::from_utf8(buffer).map_err(Into::into)
+        } else {
+            Err(anyhow!("Can’t represent {:?} as str", self.value))
```
surma, line 213:
```suggestion
            Err(anyhow!("Value {:?} is not a str", self.value))
```

crates/quickjs-wasm-rs/src/lib.rs
saulecabrera: It does, thanks!

saulecabrera (approved):

jeffcharles:
Please update the CHANGELOG file for `quickjs-wasm-rs` and update the version of the crate to `3.1.0-alpha.1` since this adds meaningful new features.

crates/quickjs-wasm-rs/src/js_binding/value.rs
surma: Removing this from the PR

surma:
@jeffcharles @saulecabrera I think this is good to go

Issue #608, opened 2024-02-27
[quickjs-wasm-rs] Should the closure passed to `wrap_callback` return `JSValueRef`?


surma:
Currently, the closure passed to `wrap_callback` is `impl FnMut(&Self, JSValueRef<'_>, &[JSValueRef<'_>]) -> Result<JSValue> + 'static`. 

This has worked well so far, however, this makes it impossible to return an already existing value and maintain referential equality. For example:

```rust
let ctx = JSContextRef::default();
// This is supposed to be a Rust-implementation of `x => x`.
let identity_func = ctx.wrap_callback(|ctx, _this, args| {
    let undef = ctx.undefined_value()?;
    // As far as I can tell, `from_qjs_value` is the only path from `JSValueRef` 
    // to `JSValue`, which deep-copies the value into a context-independent data structure.
    from_qjs_value(args.get(0).cloned().unwrap_or(undef))
})?;
let object = ctx.object_value()?;
let result = identity_func.call(&ctx.undefined_value()?, &[func.clone()])?;
// Will fail as these are different objects now.
assert_eq!(Into::<u64>::into(result), Into::<u64>::into(object));
```

As `JSValue` is a context-independent representation of a JS value, would it make sense for a callback to always return a context-specific value, which can be easily created from a given `JSValue` using `to_qjs_value()`?

(Side-note: Would be great to also implement `PartialEq` for `JSValueRef`.)

saulecabrera:
I _think_ we did it in this way initially assuming that all use-cases would want to return high-level Rust types from the callback, but taking a close look what you're suggesting, it doesn't seem like the right assumption and makes sense to me also because `wrap_callback` immediately converts the `JSValue` to a `JSValueRef` in which case if you want to pass an existing `JSValueRef` the process feels somewhat inefficient. I think that we could make it so that it works for both use cases, and let the caller decide what return type they are after depending on the use-case.

saulecabrera:
Now that https://github.com/bytecodealliance/javy/pull/618 is merged, this is no longer an issue. I'll go ahead and close this issue.

PR #618, opened 2024-03-19
Switch to `rquickjs`

saulecabrera:
## Description of the change

Historically, the Javy project has relied on its own QuickJS bindings in order to consume QuickJS' functionality. When the project started, there were very limited options for bindings that could work for our use-case and more importantly, bindings that would compile to `wasm32-wasi`. That said, bindings under this project are "good enough" for Javy, but they are probably not the most ergonomic for use-cases that require more complex interactions with QuickJS.

The Rust <> QuickJS landscape has changed in the past few years, which means that it's probably a good time to reconsider our initial decisions.

This PR proposes migrating to `rquickjs` and dropping support for `quickjs-wasm-rs` and `quickjs-wasm-sys`. The `rquickjs` bindings are more ergonomic and enable more natural interactions between Rust and QuickJS' API. 
 
With the adoption of `rquickjs`, we'd be able to address issues like: https://github.com/bytecodealliance/javy/issues/605, https://github.com/bytecodealliance/javy/issues/608, https://github.com/bytecodealliance/javy/pull/617 and https://github.com/bytecodealliance/javy/pull/611. 

This change is divided in a series of commits, but it's better reviewed as a whole, given that I didn't put enough thought to make each commit self-contained.

In general, this change doesn't introduce any functional changes, dropping support for `quickjs-wasm-rs` is the only breaking change introduced by this PR. The only exception is that I fixed a bug in our number deserialization process, in which the implementation was not respecting the `MIN_SAFE_INTEGER..=MAX_SAFE_INTEGER` range. 

This change is already big as-is; I'm planning on creating follow-up PRs for:

- [ ] Updating the documentation under `docs`
- [ ] Update the READMEs for `quickjs-wasm-rs` and `quickjs-wasm-sys` with a deprecation notice.

## Checklist

### DIFF
--- a/crates/apis/src/console/mod.rs
+++ b/crates/apis/src/console/mod.rs
@@ -14,7 +14,6 @@ pub use config::LogStream;
 
 mod config;
 
-// TODO: #[derive(Default)]
 pub(super) struct Console {}
 
 impl Console {
@@ -33,17 +32,11 @@ impl JSApiSet for Console {
     }
 }
 
-fn register_console<'js, T, U>(
-    context: &Context,
-    mut log_stream: T,
-    mut error_stream: U,
-) -> Result<()>
+fn register_console<T, U>(context: &Context, mut log_stream: T, mut error_stream: U) -> Result<()>
 where
     T: Write + 'static,
     U: Write + 'static,
 {
-    // TODO: Revisit the callback signatures, there's a possibility that we can
-    // actually convert from anyhow::Error to quickjs::Error.
     context.with(|this| {
         let globals = this.globals();
         let console = Object::new(this.clone())?;
--- a/crates/apis/src/console/mod.rs
+++ b/crates/apis/src/console/mod.rs
@@ -1,8 +1,8 @@
 use std::io::Write;
 
-use anyhow::Result;
+use anyhow::{Error, Result};
 use javy::{
-    quickjs::{JSContextRef, JSValue, JSValueRef},
+    quickjs::{prelude::Rest, Context, Ctx, Function, Object, Rest, Value},
     Runtime,
 };
 
@@ -31,21 +31,35 @@ impl JSApiSet for Console {
     }
 }
 
-fn register_console<T, U>(context: &JSContextRef, log_stream: T, error_stream: U) -> Result<()>
+fn register_console<'js, T, U>(context: &Context, log_stream: T, error_stream: U) -> Result<()>
 where
     T: Write + 'static,
     U: Write + 'static,
 {
-    let console_log_callback = context.wrap_callback(console_log_to(log_stream))?;
-    let console_error_callback = context.wrap_callback(console_log_to(error_stream))?;
-    let console_object = context.object_value()?;
-    console_object.set_property("log", console_log_callback)?;
-    console_object.set_property("error", console_error_callback)?;
-    context
-        .global_object()?
-        .set_property("console", console_object)?;
+    context.with(|cx| {
+        let globals = cx.globals();
+        let console = Object::new(cx)?;
+        console.set(
+            "log",
+            Function::new(cx, |cx: Ctx<'js>, args: Rest<Value<'js>>| {
+                Value::new_undefined(cx)
+            }),
+        )?;
+
+        console.set(
+            "error",
+            Function::new(cx, |cx: Ctx<'js>, args: Rest<Value<'js>>| {
+                Value::new_undefined(cx)
+            }),
+        )?;
+
+        globals.set("console", console)?;
+        Ok::<_, Error>(())
+    });
     Ok(())
 }
+
+fn log<'js, T>(mut stream: T, cx: Ctx<'js>, args: &[Value<'js>]) -> Result<Value<'js>> {}
 
 fn console_log_to<T>(
     mut stream: T,
@@ -53,7 +67,7 @@ fn console_log_to<T>(
 where
     T: Write + 'static,
 {
-    move |_ctx: &JSContextRef, _this: JSValueRef, args: &[JSValueRef]| {
+    move |_ctx: Cx<'_>, _this: JSValueRef, args: &[JSValueRef]| {
         // Write full string to in-memory destination before writing to stream since each write call to the stream
         // will invoke a hostcall.
         let mut log_line = String::new();
--- a/crates/apis/src/console/mod.rs
+++ b/crates/apis/src/console/mod.rs
@@ -2,11 +2,12 @@ use std::io::Write;
 
 use anyhow::{Error, Result};
 use javy::{
+    hold, hold_and_release, print,
     quickjs::{prelude::MutFn, Context, Function, Object, Value},
-    Runtime,
+    to_js_error, Args, Runtime,
 };
 
-use crate::{print, APIConfig, Args, JSApiSet};
+use crate::{APIConfig, JSApiSet};
 
 pub(super) use config::ConsoleConfig;
 pub use config::LogStream;
@@ -52,7 +53,8 @@ where
             Function::new(
                 this.clone(),
                 MutFn::new(move |cx, args| {
-                    log(Args::hold(cx, args), &mut log_stream).expect("console.log to succeed")
+                    let (cx, args) = hold_and_release!(cx, args);
+                    log(hold!(cx.clone(), args), &mut log_stream).map_err(|e| to_js_error(cx, e))
                 }),
             )?,
         )?;
@@ -60,9 +62,10 @@ where
         console.set(
             "error",
             Function::new(
-                this,
+                this.clone(),
                 MutFn::new(move |cx, args| {
-                    log(Args::hold(cx, args), &mut error_stream).expect("console.error to succeed")
+                    let (cx, args) = hold_and_release!(cx, args);
+                    log(hold!(cx.clone(), args), &mut error_stream).map_err(|e| to_js_error(cx, e))
                 }),
             )?,
         )?;
--- a/crates/apis/src/console/mod.rs
+++ b/crates/apis/src/console/mod.rs
@@ -2,11 +2,14 @@ use std::io::Write;
 
 use anyhow::{Error, Result};
 use javy::{
-    quickjs::{prelude::Rest, Context, Ctx, Function, Object, Rest, Value},
+    quickjs::{
+        prelude::{MutFn, Rest},
+        Context, Ctx, Function, Object, Value,
+    },
     Runtime,
 };
 
-use crate::{APIConfig, JSApiSet};
+use crate::{print, APIConfig, JSApiSet};
 
 pub(super) use config::ConsoleConfig;
 pub use config::LogStream;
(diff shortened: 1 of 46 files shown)


surma (commented):
I know you didn’t solicit a review and I didn’t really do a full review, but I wanted to take a look where you are at and left 2 comments :3

crates/apis/src/console/mod.rs
```diff
@@ -31,50 +33,71 @@ impl JSApiSet for Console {
...
+                MutFn::new(move |cx, args| {
+                    let (cx, args) = hold_and_release!(cx, args);
+                    log(hold!(cx.clone(), args), &mut error_stream).map_err(|e| to_js_error(cx, e))
+                }),
+            )?,
+        )?;
+
+        globals.set("console", console)?;
+        Ok::<_, Error>(())
+    })?;
```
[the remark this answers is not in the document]
saulecabrera: I was going to ask your review anyway!

This comment is stale by now; previously here I had a `.expect()`,  but now each callback, correctly maps `anyhow::Error` to `JSError`.  Does that address your comment?

crates/javy/src/lib.rs
```diff
@@ -40,15 +40,115 @@
...
+        }
+        Type::Float => {
+            let f = val.as_float().unwrap();
+            write!(sink, "{}", f).map_err(Into::into)
+        }
+        Type::String => {
+            let s = val.as_string().unwrap();
+            write!(sink, "{}", s.to_string()?).map_err(Into::into)
+        }
+        Type::Array => {
```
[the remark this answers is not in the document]
saulecabrera: I guess it could, but I think that part of a bigger conversation though, given that there are many conventions that we could follow, for now I've kept it similar to the previous implementation:

```js
console.log([1, 2, 4]);
```


```
❯ npx javy-cli compile index.js -o index.wasm
❯ wasmtime run --invoke=_start index.wasm
1,2,4
```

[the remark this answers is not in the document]
saulecabrera: Not saying that we shouldn't do it, mostly that it's probably outside of the scope of this change.

crates/apis/src/console/mod.rs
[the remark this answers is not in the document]
makrisoft: I think because we need access to the context, we need to map the error at that level. I was trying to think of a way to dry this up, if there was an easy way (without pulling in a crate) to create a curried function with cx, then we could just pass that if it supported FunOnce I think......

The other thing I was thinking was if we could push the to_js_error into log, but that sucks since there are multiple exit points and we would be repeating map_err's again.....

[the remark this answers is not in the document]
saulecabrera: I did thought about creating a wrapper that would allow you to automatically do the error handling (similar to currying, yeah) but I didn't invest much time on it. Another option is to create a macro that handles all this boilerplate. I'd like to land this one as is since it's fairly large already and I can follow-up on this specific piece after the fact.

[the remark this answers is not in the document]
surma: Oh, so I only left the comment because of the `TODO`.

Before creating a wrapper type I’d recommend going with `thiserror`. In general I prefer using that for libraries and only use `anyhow` for leaves (apps, servers etc). That would allow you to convert _from_ different error types to your own, and then you could just implement the `IntoJs` trait on it to make rquickjs be able to deal with it :)

saulecabrera:
Ok, so all tests are passing here. I'm working on migrating our json/msgpack ser/de and that should be the last step (aside from vetting the new dependencies) to get this in a review-ready state.

crates/javy/src/lib.rs
```diff
@@ -40,15 +41,173 @@
...
+            let inner = val.as_array().unwrap();
+            for e in inner.iter() {
+                print(&e?, sink)?
+            }
+            Ok(())
+        }
+        Type::Object => write!(sink, "[object Object]").map_err(Into::into),
+        // TODO: Implement the rest.
+        x => unimplemented!("{x}"),
```
saulecabrera, line 97: This matches our previous implementation, as far as I can tell, but I'll do more extensive testing and update if necessary.

crates/apis/src/stream_io/io.js
```diff
@@ -19,7 +19,7 @@
       }
       return __javy_io_writeSync(
         fd,
-        data.buffer,
+        data,
```
saulecabrera, line 22: Changed this to make it easier to work with the JS API from the Rust side. In theory we could change the entire API so that Rust only receives `data`, given that we're certain at this point that this is an `TypedArray` and therefore we can derive the `byteOffset` and `byteLength`. This can be done in a follow-up though.

crates/apis/src/stream_io/mod.rs
```diff
@@ -1,68 +1,163 @@
...
+    let (cx, args) = args.release();
+    let (fd, data, offset, length) = extract_args(&args, "Javy.IO.writeSync")?;
+    let fd = match fd
+        .as_int()
+        .ok_or_else(|| anyhow!("File descriptor must be a number"))?
+    {
+        1 => Fd::Stdout,
+        2 => Fd::Stderr,
+        x => anyhow::bail!(
+            "Wrong file descriptor: {}. Only stdin(1) and stderr(2) are supported",
```
jeffcharles, line 55:
```suggestion
            "Unsupported file descriptor: {x}. Only stdout(1) and stderr(2) are supported"
```

crates/apis/src/stream_io/mod.rs
```diff
@@ -1,68 +1,163 @@
...
+fn read(args: Args<'_>) -> Result<Value<'_>> {
+    let (cx, args) = args.release();
+    let (fd, data, offset, length) = extract_args(&args, "Javy.IO.readSync")?;
+
+    let mut fd: Stdin = match fd
+        .as_int()
+        .ok_or_else(|| anyhow!("File descriptor must be a number"))?
+    {
+        0 => std::io::stdin(),
+        x => anyhow::bail!("Wrong file descriptor: {}. Only stdin(1) is supported", x),
```
jeffcharles, line 101:
```suggestion
        x => anyhow::bail!("Unsupported file descriptor: {x}. Only stdin(0) is supported"),
```

crates/core/src/lib.rs
```diff
@@ -37,15 +45,15 @@ pub unsafe extern "C" fn compile_src(js_src_ptr: *const u8, js_src_len: usize) -
...
     let bytecode = runtime
-        .context()
-        .compile_module(FUNCTION_MODULE_NAME, js_src)
+        .compile_to_bytecode(FUNCTION_MODULE_NAME, js_src)
         .unwrap();
-    let bytecode_len = bytecode.len();
+
     // We need the bytecode buffer to live longer than this function so it can be read from memory
-    let bytecode_ptr = Box::leak(bytecode.into_boxed_slice()).as_ptr();
+    let bytecode_ptr = Box::leak(bytecode.clone().into_boxed_slice()).as_ptr();
```
jeffcharles, line 54: Nit: I'm interested in why this was changed to introduce a clone. It works so it's fine.

crates/javy/src/runtime.rs
```diff
@@ -1,59 +1,86 @@
...
+/// [rquickjs] APIs.
 pub struct Runtime {
-    context: JSContextRef,
+    /// The QuickJS context.
+    context: Context,
+    /// The inner QuickJS runtime representation.
+    // We use `ManuallyDrop` to avoid incurring in the cost of dropping the
+    // [rquickjs::Runtime] and its associated objects, which takes a substantial
+    // time.
+    // This assumes that Javy is used for short-lived programs were the host
```
jeffcharles, line 21:
```suggestion
    // This assumes that Javy is used for short-lived programs where the host
```

crates/javy/src/runtime.rs
```diff
@@ -1,59 +1,86 @@
...
+/// Javy's [Runtime] holds a [rquickjs::Runtime] and [rquickjs::Context],
+/// and provides accessors these two propoerties which enable working with
+/// [rquickjs] APIs.
 pub struct Runtime {
-    context: JSContextRef,
+    /// The QuickJS context.
+    context: Context,
+    /// The inner QuickJS runtime representation.
+    // We use `ManuallyDrop` to avoid incurring in the cost of dropping the
+    // [rquickjs::Runtime] and its associated objects, which takes a substantial
```
jeffcharles, line 19: Nit: IMO we should be using a monospaced font for types in documentation (that is, wrap the type name in backticks) to improve readability.

jeffcharles (commented):
We should update the READMEs for `quickjs-wasm-sys` and `quickjs-wasm-rs` to indicate they are deprecated in favour of `rquickjs`. Would that make sense to do as part of this PR or in a follow up one?

crates/javy/src/runtime.rs
saulecabrera: I think here it's fine to use backticks according to your suggestion given that this is not a doc comment. In places where we have proper doc comments, I do prefer to use `[T]` or `[`T`]`, to enable [links to items by name](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html), as it makes navigating to those items easier, both from code if you have an LSP or from docs.rs. For these cases I find `[T]`more readable than `[`T`]`.

crates/core/src/lib.rs
saulecabrera: Good catch, yeah. I've restored the previous implementation to prevent that clone. The clone was required because was I was using `bytecode.len()` after calling `into_boxed_slice`.

saulecabrera:
> We should update the READMEs for `quickjs-wasm-sys` and `quickjs-wasm-rs` to indicate they are deprecated in favour of `rquickjs`. Would that make sense to do as part of this PR or in a follow up one?

I'll do that as a follow-up. I added an entry in the PR description.

crates/javy/src/runtime.rs
jeffcharles: I meant to say we should have the type name within backticks **and** square brackets in doc comments. That is, write them as ```[`T`]```. There's a number of places with new doc comments where we're not doing that.

You can see the difference in how the crate documentation is rendered:

Screenshot 2024-04-17 at 9 51 24 AM

The `Args` doc comment (introduced in this PR) doesn't use backticks inside its square brackets for its links to `Ctx` and `Value` but `Config`'s doc comment does for `Runtime` and IMO, the `Config` doc comment looks more readable to me.

It's also not a huge deal to change later so I don't want to hold up the PR on them but it'd be nice to not have to for doc comments added by this PR.

saulecabrera: Oh got it, yeah in that case I agree. I wrongly assumed that in both cases (`[]` and ```[``]```) this would end up  highlighted as code. I don't want to introduce more maintenance here, so I can definitely do a pass over all the new comments and make sure they have backticks.

saulecabrera:
Oops meant to push https://github.com/bytecodealliance/javy/pull/618/commits/5fde19c7316d693bb895e0b169a038622a83e657, into a copy branch to do some testing -- I'll revert it.

makrisoft (approved):

crates/cli/tests/integration_test.rs
saulecabrera, line 1: It's worth noting that v0.6.0 updates to the latest QuickJS version, which returns a promise when evaluating QuickJS bytecode, which is responsible for a small fuel increase in some cases (observed ~12k units of fuel on average). But in other cases, the reduction in fuel is also considerable.

supply-chain/config.toml
```diff
@@ -34,6 +34,18 @@ audit-as-crates-io = false
 [policy.quickjs-wasm-sys]
 audit-as-crates-io = false
```
saulecabrera, line 36: Temporary until we're off the fork; I intended to add a comment but `cargo vet` didn't liked it.

crates/core/src/execution.rs
```diff
@@ -1,20 +1,40 @@
...
     runtime
         .context()
-        .with(|this| Module::instantiate_read_object(this, bytecode).map(|_| ()))
+        .with(|this| {
+            let module = unsafe { Module::load(this.clone(), bytecode)? };
+            let (_, promise) = module.eval()?;
+
+            if cfg!(feature = "experimental_event_loop") {
+                // If the experimental event loop is enabled, trigger it.
+                promise.finish::<Value>().map(|_| ())
```
jeffcharles, line 29: What does the `.map(|_| ())` do? Could we just do `promise.finish::<Value>();` if we want to swallow the return value?

saulecabrera: That will introduce incompatible return types for if/else: `()` vs `Result<()>`

redoC-A2k:
Thanks a lot for doing switching to rquickjs work .Till when can I expect the new release on github roughly ?

saulecabrera:
Mid June is my estimate: we're doing more extensive testing and working on other improvements before releasing the next version.

sandstrom:
I'm trying to figure out which version of QuickJS that the Javy's master branch is using.

Is it 2024-01-13 or 2023-12-09?

https://bellard.org/quickjs/

saulecabrera:
https://github.com/bellard/quickjs/blob/3b45d155c77bbdfe9177b1e03db830d2aff0b2a8/VERSION -- we currently rely on `rquickjs` version `0.6.1`, which points to the link above, which defines version 2024-02-14

sandstrom:
Thanks @saulecabrera!

PR #639, opened 2024-05-03
Skip serializing object entries w/ undefined value

jbourassa:
## Description of the change

Improve Javy's Value serializer compatibility with `JSON.stringify`:
- For Arrays, nullify non-JSONable values
- For Objects, skip entries with non-JSONable values

Encoding a QuickJS value to JSON using the transcoder offers a nice performance gain over QuickJS's `JSON.stringify`, but differs in behaviour. This PR should partially close the gap.

```
$ echo "console.log(JSON.stringify({foo: undefined}))" | node
{}
```

Without this change, transcoding the above `{foo: undefined}` to a JSON string would result in `{"foo":null}`.

This PR fixes that by skipping skipping the deserialization of entries with `undefined`. **This is a lossy conversion** given serde data model doesn't know about `undefined`,  and both the serde `Unit` and `None` data types are serialized as `null` in JSON.

Similar logic is applied to Array values, but in that case the values are nullified instead of skipped.

## Checklist

- ~[ ] I've updated the relevant crate versions if necessary. [Versioning policy for library crates](https://github.com/bytecodealliance/javy/blob/main/docs/contributing.md#versioning-for-library-crates)~
- ~[ ] I've updated documentation including crate documentation if necessary.~

### DIFF
--- a/crates/javy/CHANGELOG.md
+++ b/crates/javy/CHANGELOG.md
@@ -10,7 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Introduce `rquickjs` to interface with QuickJS instead of `quickjs-wasm-rs`;
-  this version no longer includes re-exports from `quickjs-wasm-rs`. 
+  this version no longer includes re-exports from `quickjs-wasm-rs`.
+- `javy::serde::de::Deserializer` should now match `JSON.stringify`: non-JSON
+  primitives are skipped in Objects and nullified in Arrays.
 
 ## [2.2.0] - 2024-01-31
 
--- a/crates/javy/src/serde/de.rs
+++ b/crates/javy/src/serde/de.rs
@@ -3,6 +3,7 @@ use crate::serde::err::{Error, Result};
 use crate::serde::{MAX_SAFE_INTEGER, MIN_SAFE_INTEGER};
 use crate::{from_js_error, to_string_lossy};
 use anyhow::anyhow;
+use rquickjs::Null;
 use serde::de::{self, Error as SerError};
 use serde::forward_to_deserialize_any;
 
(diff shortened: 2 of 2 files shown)


saulecabrera (commented):
This change looks good to me, but it also got me thinking about other things:

- I _think_ we should also check if values are functions and values are symbols? That's what the official [JS JSON spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) states: 
> [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined), [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function), and [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) values are not valid JSON values. If any such values are encountered during conversion, they are either omitted (when found in an object) or changed to [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null) (when found in an array). JSON.stringify() can return undefined when passing in "pure" values like JSON.stringify(() => {}) or JSON.stringify(undefined).

- I'm wondering how this affects the messagepack implementation, if anything at all, given that this deserializer is used there too. I don't think that the messagepack crate is used at all in general, maybe another sensible option is to deprecate it.

jbourassa:
> I think we should also check if values are functions and values are symbols?

Good point, I can throw those in too. Hopefully it has limited impact on performance.

>I'm wondering how this affects the messagepack implementation, if anything at all, given that this deserializer is used there too.

It will, but arguably that's a feature? I'd expect the msgpack implementation to be _exactly_ like `JSON.stringify`, but msgpack encoded.

jbourassa:
@saulecabrera I've now applied your suggestion of checking for more types. Sorry for the delay!

saulecabrera (approved):
Looks great, thanks!

PR #640, opened 2024-05-06
docs: Improve READMEs for `apis` and `javy` crates

saulecabrera:
## Description of the change

In preparation to adding more APIs I decided to do a very light cleanup of some of the documentation in this crate. I opted to remove the documentation in the README to reduce the overhead of having to maintain the same documentation in multiple places.

## Checklist

### DIFF
--- a/crates/apis/src/lib.rs
+++ b/crates/apis/src/lib.rs
@@ -3,25 +3,23 @@
 //! APIs are enabled through cargo features.
 //!
 //! Example usage:
-//! ```
-//! # use anyhow::{anyhow, Error, Result};
-//! use javy::{quickjs::JSValue, Runtime};
-//! use javy_apis::RuntimeExt;
+//! ```rust
 //!
-//! let runtime = Runtime::new_with_defaults()?;
-//! let context = runtime.context();
-//! context.global_object()?.set_property(
-//!    "print",
-//!    context.wrap_callback(move |_ctx, _this, args| {
-//!        let str = args
-//!            .first()
-//!            .ok_or(anyhow!("Need to pass an argument"))?
-//!            .to_string();
-//!        println!("{str}");
-//!        Ok(JSValue::Undefined)
-//!    })?,
-//! )?;
-//! context.eval_global("hello.js", "print('hello!');")?;
+//! //With the `console` feature enabled.
+//! use javy::{Runtime, from_js_error};
+//! use javy_apis::RuntimeExt;
+//! use anyhow::Result;
+//!
+//! fn main() -> Result<()> {
+//!     let runtime = Runtime::new_with_defaults()?;
+//!     let context = runtime.context();
+//!     context.with(|cx| {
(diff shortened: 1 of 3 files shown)


crates/apis/src/lib.rs
```diff
@@ -30,17 +30,18 @@
...
+//! * `text_encoding`:  Registers implementations of `TextEncoder` and `TextDecoder`.
+//! * `random`: Overrides the implementation of `Math.random` to one that seeds
+//! the RNG on first call to `Math.random`. This is helpful to enable when using
+//! using a tool like Wizer to snapshot a [`Runtime`] so that the output of
+//! `Math.random` relies on the WASI context used at runtime and not the WASI
+//! context used when Wizening. Enabling this feature will increase the size of
+//! the Wasm module that includes the Javy Runtime and will introduce an
+//! additional hostcall invocation when `Math.random` is invoked for the first
+//! time.
+//! * `stream_io`: Adds the implementation of `Javy.IO.readSync` and `Javy.IO.writeSync`.
```
jeffcharles, line 44:
```suggestion
//! * `stream_io`: Adds an implementation of `Javy.IO.readSync` and `Javy.IO.writeSync`.
```

jeffcharles (commented):
Content changes look fine to me!

Could we do something like `#[doc = include_str!("../README.md")]` to generate the module level documentation from the README? That should get rid of the duplication. I like having some sort of documentation in the README since that's what shows up on the crates.io page for the crate. _I_ understand developers can click through the documentation link but most crates have documentation in their READMEs so I think it's somewhat expected for READMEs.

saulecabrera:
I had the intention to do something along those lines, I decided that it's probably not worth the trouble because we probably don't want to include everything in the README in the crate-level documentation e.g. the `## Publishing to crates.io` section and other README-specific sections that we might add in the future. Personally I find it more valuable to have great crate-level docs (showing up in docs.rs) than having some content in the README in crates.io.

saulecabrera:
On a second thought, bringing just the example at least might be reasonable to do, I suspect that one won't need to be update as frequently.

saulecabrera:
Added the example, and I also realized just now that the example in `lib.rs` contained a snippet using the old bindings, so I also fixed that.

saulecabrera:
Ok so got a bit inspired and improved the READMEs across the board they not only include a direct link to docs.rs but also the badges to crates.io and docs.rs. With these changes I don't think there's a chance of confusion around how to navigate to the crate-level documentation.

crates/javy/README.md
```diff
@@ -33,14 +43,6 @@ context.with(|cx| {
 });
 ```
 
-Create a `Runtime` and use the reference returned by `context()` to add functions and evaluate source code.
```
saulecabrera, line 36: This is an instance of out-of-date documentation. The crate-level docs have a slightly different shape.

jeffcharles (approved):
This looks good to me! Thanks for keeping an example of the usage in the README, I think it helps provide context for when and why someone may want to use these crates.

PR #644, opened 2024-05-08
Leverage QuickJS's `JS_ToString` in logging

jbourassa:
## The problem

Javy crashes when `console.log`ing certain JavaScript values. 

## Description of the change

Use QuickJS's `JS_ToString` implementation for `console.{log,error}` instead of rolling our own; reverting to what Javy 2.x did. 

This change also adds new serialization test cases to capture the current behaviour. Symbols are excluded because they crashed on Javy 2.x and throw an exception with this branch (from this [QuickJS line](https://github.com/bellard/quickjs/blob/01454caf78bcc928dfab1ca62a551fe92ada4d4c/quickjs.c#L11724)). See [this repro script and outputs](https://gist.github.com/jbourassa/ede119af314be619390636f08f2b337b).

## Tradeoffs

I suspect this is slower because we'll allocate Strings for any values. At the same time, we get to leverage QuickJS' implementation just like we did prior to the introduction of the rquickjs bindings.

The alternative is to re-implement printing each types, thus choosing a string representation for each type. If we want to go that route, we have to choose wether we want to follow what other engines do (more work, slower), or stick with Javy's current minimal approach.

## Checklist

- ~[ ] I've updated the relevant CHANGELOG files if necessary. Changes to `javy-cli` and `javy-core` do not require updating CHANGELOG files.~
- ~[ ] I've updated the relevant crate versions if necessary. [Versioning policy for library crates](https://github.com/bytecodealliance/javy/blob/main/docs/contributing.md#versioning-for-library-crates)~
- ~[ ] I've updated documentation including crate documentation if necessary.~

### DIFF
--- a/crates/apis/src/console/mod.rs
+++ b/crates/apis/src/console/mod.rs
@@ -73,17 +73,29 @@ where
 
 fn log<'js, T: Write>(args: Args<'js>, stream: &mut T) -> Result<Value<'js>> {
     let (ctx, args) = args.release();
-    for (i, arg) in args.iter().enumerate() {
-        let stringified =
-            <convert::Coerced<JSString>>::from_js(&ctx, arg.clone()).map(|string| {
-                string
-                    .to_string()
-                    .unwrap_or_else(|e| to_string_lossy(&ctx, &string.0, e))
-            })?;
+    for (i, arg) in args.into_inner().into_iter().enumerate() {
         if i != 0 {
             write!(stream, " ")?;
         }
-        write!(stream, "{stringified}")?;
+
+        if let Some(symbol) = arg.as_symbol() {
+            if let Some(description) = symbol.description()?.into_string() {
+                let description = description
+                    .to_string()
+                    .unwrap_or_else(|e| to_string_lossy(&ctx, &description, e));
+                write!(stream, "Symbol({description})")?;
+            } else {
+                write!(stream, "Symbol()")?;
+            }
+        } else {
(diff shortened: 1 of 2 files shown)


saulecabrera (commented):
In general looks good to me! I believe that printing strings has the potential to fail in this case, because rquickjs doesn't fully handle the conversion between JavaScript string encoding and Rust (utf-8) encoding. We have https://github.com/bytecodealliance/javy/blob/main/crates/javy/src/lib.rs#L164 to deal with that case. Do you mind verifying that this edge case works as expected?

saulecabrera:
Also, it's not entirely clear to me why Symbols are excluded in QuickJS itself. By checking the behavior in other engines, I can confirm that they are correctly stringified as `Symbol(v)` where `v` is the [symbol's description field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/description). Could we perhaps special case symbols so the behavior matches?

jbourassa:
✅ Symbol special-casing

> Also, it's not entirely clear to me why Symbols are excluded in QuickJS itself.

Same. It looks intentional though, but probably not in the case of `JS_ToString` -- it probably makes sense for other scenarios.

-----

✅ wtf8 support, the added test is a little crude but I think it works?

LMK what you think!

saulecabrera (approved):
Looks great, thanks for fixing this.

Issue #628, opened 2024-04-13
Allow specifying which file descriptor is used in console.{log,err}  from the CLI


Niikelion:
Operating system: windows + wsl2-ubuntu, windows
Processor architecture: x86_64
Rust version: rustup 1.27.0
Javy version: javy 1.4.0

## Problem
After compiling `console.log("test")` into `.wasm` file and executing it with wasmtime I get output on stderr instead of stdout.
I verified that its not an wasmtime problem by compiling and running simple c program with `printf("test\n")`. Testing was done in windows and wsl2 with ubuntu on windows and in happens in both cases.

saulecabrera:
It indeed does. I believe we should make this behavior configurable from the CLI level too, because currently it's not standard. For some context, some use-cases require tight control over which stream is used for `console.{log,err}`, however, in this case the non-standard behavior has "leaked" as the standard one.

saulecabrera:
I'll switch the label and title of the issue to reflect that we want to make this behavior configurable.

saulecabrera:
In the latest main, it's now possible to override this behavior, by doing `javy build index.js -J redirect-stdout-to-stderr=n -o index.wasm`

PR #646, opened 2024-05-09
Refactor the structure of the `Config`/APIs

saulecabrera:
This change primarily rewrites the current APIs using "Intrinsics" as defined by rquickjs.

This change has several objectives:

- Refactor the structure of the Config and its relationship with the Runtime to pave the way for a truly extensible and configurable engine. The way Intrinsics are exposed in rquickjs simplifies (i) enabling a granular configuration experience of the native intrinsics, which was not entirely clear how to achieve before, and (ii) adding custom APIs as intrinsics.
 
- Enhance how our configuration operates: the previous model assumed an almost static configuration setup, allowing only limited, non-straightforward customizability from the CLI in the future. This was mostly modeled per API and not as a global aspect of the runtime. Our recommended approach for users needing functionality beyond what was provided was to fork or recreate their own version of the CLI/APIs. A telling sign of this issue is that the configuration parameters were almost never used in any of the APIs. This PR paves the way for greater extensibility, which will be facilitated by threading configuration options from the CLI in a follow-up PR; this method will also allow for opting-in to non-standard functionality via a CLI flag, helping us avoid issues like the one noted here: https://github.com/bytecodealliance/javy/issues/628.

Along the way,  I had to make several changes to make all this work; the most relevant ones  are: (i)  deprecating the `javy-apis` crate (ii) enabling all the APIs by default and (iii) reworking a bit the `Console` implementation. 

The reasoning to deprecate the `javy-apis` crate is mostly around: (i) the fact that we're revamping our extensilbity model in the near fuiture and (ii)  the complexity around cyclical dependencies:  `javy` exposes rquickjs and given that the cofiguration is tied to the runtime, `javy` needed to depend on `javy-apis` but the other way around was also true. I decided not to spend too much cycles on this mostly given point (i) above. 

There are still things to follow-up on after this PR, namely 

- [ ] Update all the documentation under  `docs` 
- [ ] A PR to thread the engine configuration from the CLI. **NB that the engine behaviour is not  altered in this PR.**

## Checklist


saulecabrera:
cc/ @surma since we chatted about using Intrinsics, I decided to give this a try as part of adding a faster JSON implementation.

crates/core/src/runtime.rs
```diff
@@ -1,9 +1,12 @@
...
 pub(crate) fn new_runtime() -> Result<Runtime> {
-    let mut api_config = APIConfig::default();
-    api_config.log_stream(LogStream::StdErr);
-    Runtime::new_with_apis(Config::default(), api_config)
+    let mut config = Config::default();
+    let config = config
+        .text_encoding(true)
+        .redirect_stdout_to_stderr(true)
+        .javy_stream_io(true);
+
```
saulecabrera, line 10: This will be made configurable in a later PR.

crates/core/Cargo.toml
```diff
@@ -16,7 +16,7 @@ crate-type = ["cdylib"]
 [dependencies]
 anyhow = { workspace = true }
 javy = { workspace = true, features = ["export_alloc_fns"] }
-javy-apis = { path = "../apis", features = ["console", "text_encoding", "random", "stream_io"] }
+javy-apis = { path = "../apis" }
```
jeffcharles, line 19: I think you could just remove this if we're no longer using the crate

crates/javy/src/apis/console/mod.rs
```diff
@@ -1,73 +1,68 @@
...
+        context::Intrinsic, convert, prelude::MutFn, qjs, Ctx, FromJs, Function, Object,
+        String as JSString, Value,
     },
-    to_js_error, to_string_lossy, Args, Runtime,
+    to_js_error, to_string_lossy, Args,
 };
+use anyhow::Result;
 
-use crate::{APIConfig, JSApiSet};
+/// An implemetation of JavaScript `console` APIs.
```
jeffcharles, line 14:
```suggestion
/// An implementation of JavaScript `console` APIs.
```

crates/javy/src/apis/console/mod.rs
```diff
@@ -1,73 +1,68 @@
...
+/// This implementation is *not* standard in the sense that it redirects the output of
+/// `console.log` to stderr.
+pub(crate) struct NonStandardConsole;
 
-pub(super) use config::ConsoleConfig;
-pub use config::LogStream;
+/// An implemetation of JavaScript `console` APIs. This implementation is
+/// standard as it redirects `console.log` to stdout `console.error` to stderr.
```
jeffcharles, line 20:
```suggestion
/// standard as it redirects `console.log` to stdout and `console.error` to stderr.
```

crates/javy/src/runtime.rs
```diff
@@ -31,15 +37,96 @@ pub struct Runtime {
...
+        let context = Self::build_from_config(&rt, &config)?;
         Ok(Self { inner: rt, context })
     }
 
+    fn build_from_config(rt: &QRuntime, cfg: &Config) -> Result<ManuallyDrop<Context>> {
+        let intrinsics = &cfg.intrinsics;
+        let javy_intrinsics = &cfg.javy_intrinsics;
+        // We always set Random give that the principles around snapshotting and
+        // random are applicable to all uses of Javy, given that the use of
+        // Wizer is not optional.
```
jeffcharles, line 54: Is Wizer not being optional true in the case of someone using this crate as a library without using our CLI?

That said it's not a big deal since the outcome of calling `Math.random()` should be identical in the scenario where they choose not to Wizen.

jeffcharles (commented):
Do we plan to remove the stream I/O and console APIs into a plugin later after we've figured out a common extension mechanism?

crates/javy/src/runtime.rs
saulecabrera: Good point! I've amended the wording in the comment.

saulecabrera:
> Do we plan to remove the stream I/O and console APIs into a plugin later after we've figured out a common extension mechanism?

...and the soon-to-be-introduced `JSON` functions. That would be the ideal outcome I believe, given that they introduce non-standard functionality and it will also allow us to test our extension mechanism.

jeffcharles (approved):

PR #651, opened 2024-05-29
Improve json support

saulecabrera:
## Description of the change

This  commit introduces several improvements for JSON. Namely:

* Usage of SIMD JSON in order to speed up `JSON.parse` and using `serde_json` in order to speed up `JSON.stringify`. 
* Introduction of `Javy.JSON` functions as helpers to work with the stdin and stdout file descriptors. These are non-standard and live at the `Javy` namespace for that reason. These APIs should be considered unstable, meaning that there's no guarantee that they'll live in-tree forever, the intention is to move these away once our extensibility story is polished. 

-- 

The usage of the more performant JSON functions (based on SIMD and serde_json) *requires* enabling the QuickJS JSON intrinsic (see the config.rs module).  The reason behind this decision is to enable: a compliant JSON implementation that is optimized for the  hotpath: if extra arguments are passed to either JSON.{parse,stringify}, the native parse/stringify functions will be invoked, keeping the standard functionality. 

I believe that the best way to test this would be to run JSON 262 tests, to ensure that the implementation is fully compliant. However, this is not included in this PR. Such change is better packed on its own pull-request. I'll follow up with a change that includes a way to run JSON 262 tests.

## Checklist

### DIFF
--- a/crates/javy/src/apis/json.rs
+++ b/crates/javy/src/apis/json.rs
@@ -1,0 +1,187 @@
+//! High-performance JSON implementation for  Javy.
+//!
+//! The provided implementation is based on
+//! [simd-json](https://crates.io/crates/simd-json)
+//!
+//! The most efficient combination according to our experiments, is to use:
+//! * SIMD JSON for `JSON.parse`
+//! * Serde JSON for `JSON.stringify`
+//!
+//! It's also important to note that this implementation optimizes for the hot
+//! path:
+//! - If `JSON.parse` is invoked with the reviver argument, the native QuickJS
+//!   `JSON.parse` is invoked instead.
+//! - If `JSON.stringify` is invoked with the replacer and/or space arguments, the
+//!   native QuickJS `JSON.stringify` is invoked instead.
+//!
+//! The reason behind this decision is simple: most use-cases will hit the
+//! hotpath and doing any sort of inline processing of the parsed or stringified
+//! values is likely to void any performance benefits.
+use crate::{
+    hold, hold_and_release, json,
+    quickjs::{
+        context::Intrinsic,
+        prelude::{MutFn, Rest},
+        qjs, Ctx, Function, Object, String as JSString, Value,
+    },
+    to_js_error, val_to_string, Args,
+};
+
+use anyhow::{anyhow, Result};
+use std::{
+    io::{Read, Write},
+    ptr::NonNull,
+};
+
+/// Intrinsic to attach faster JSON.{parse/stringify} functions.
+pub struct Json;
+
+/// Intrinsic to attach functions under the `Javy.JSON` namespace.
+pub struct JavyJson;
+
+impl Intrinsic for JavyJson {
+    unsafe fn add_intrinsic(ctx: NonNull<qjs::JSContext>) {
+        register_javy_json(Ctx::from_raw(ctx)).expect("registering Javy.JSON builtins to succeed")
+    }
+}
+
+impl Intrinsic for Json {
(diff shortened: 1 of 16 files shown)


crates/javy/src/config.rs
```diff
@@ -161,10 +170,62 @@ impl Config {
...
+    }
+
     /// Enables whether the output of console.log will be redirected to
     /// `stderr`.
     pub fn redirect_stdout_to_stderr(&mut self, enable: bool) -> &mut Self {
         self.redirect_stdout_to_stderr = enable;
         self
     }
+
+    /// Whether to override the implemenation of JSON.parse and JSON.stringify
```
jeffcharles, line 189:
```suggestion
    /// Whether to override the implementation of JSON.parse and JSON.stringify
```

crates/javy/src/json.rs
```diff
@@ -3,16 +3,16 @@ use crate::serde::{de::Deserializer, ser::Serializer};
 use anyhow::Result;
 
 /// Transcodes a byte slice containing a JSON encoded payload into a [Value].
-pub fn transcode_input<'js>(context: Ctx<'js>, bytes: &[u8]) -> Result<Value<'js>> {
-    let mut deserializer = serde_json::Deserializer::from_slice(bytes);
+pub fn parse<'js>(context: Ctx<'js>, bytes: &mut [u8]) -> Result<Value<'js>> {
+    let mut deserializer = simd_json::Deserializer::from_slice(bytes)?;
```
jeffcharles, line 7: Would this need to be gated by a check for whether the feature that includes the `simd_json` dependency is enabled or not?

crates/javy/src/config.rs
```diff
@@ -161,10 +170,62 @@ impl Config {
...
+    /// This setting requires the `JSON` intrinsic to be enabled and the `json`
+    /// crate feature to be enabled as well.
+    /// Disabled by default.
+    #[cfg(feature = "json")]
+    pub fn override_json_parse_and_stringify(&mut self, enable: bool) -> &mut Self {
+        self.override_json_parse_and_stringify = enable;
+        self
+    }
+
+    pub(crate) fn validate(self) -> Result<Self> {
```
jeffcharles, line 201: Nit: Is there a reason this can't take a reference to `self` and return that same reference?

jeffcharles (commented):
Looks good! My only real concern is whether we need a feature check or not in the `parse` method in the `javy::json` module.

jeffcharles (approved):
Got an answer out of band about the JSON transcoding function being gated by the `json` feature so we're good!

crates/javy/src/config.rs
saulecabrera: Usually these validate methods take ownership of self to force the caller to coerce `&self` or `&mut self` to `self` and avoid any unwanted usage or mutation of the config after validation. That's the reason why I opted to take in `self`

crates/javy/src/apis/json.rs
```diff
@@ -0,0 +1,187 @@
...
+    json.set("stringify", stringify)?;
+
+    Ok(())
+}
+
+fn call_json_parse(args: Args<'_>) -> Result<Value> {
+    let (this, args) = args.release();
+
+    match args.len() {
+        0 => Err(anyhow!("Error: \"undefined\" is not valid JSON")),
```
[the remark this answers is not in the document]
saulecabrera: In Chrome's console I get: 

    SyntaxError: "undefined" is not valid JSON


Which is where I took this message. Given the difference in I'll leave this as-is for now and update once I hook the JSON conformance tests, arguably they should inform what's the expected behaviour here.

PR #661, opened 2024-06-03
Add nodejs embedding example

asprouse:
## Description of the change
Adds an example of embedding a Javy-compiled WASM module in a Node.js host application. Based on work https://github.com/asprouse/javy-nodejs-example

There is not an existing example.

## Checklist


jeffcharles (approved):
Looks good to me! Thank you for contributing!

PR #662, opened 2024-06-10
Add 262 JSON tests

saulecabrera:
## Description of the change
This PR is a follow up to https://github.com/bytecodealliance/javy/pull/651
This PR is best reviewed commit by commit. 

In general, this PR introduces tests (and updates) to Javy's custom JSON.{parse,stringify} implementation to ensure compliance with the official JSON test suite.  There are a couple of tests, 3 to be exact, that are marked as ignored, refer to the comments on each test for more details.

To ensure compliance with the official JSON test suite for JavaScript. 

## Checklist


crates/javy/src/serde/ser.rs
```diff
@@ -629,24 +631,6 @@ mod tests {
         Ok(())
     }
 
-    #[test]
```
saulecabrera, line 632: Context: https://github.com/bytecodealliance/javy/pull/662/commits/4cc80e7afe35f5a2431e29d817bf22ccf3b99e99

crates/javy/src/apis/json.rs
```diff
@@ -52,20 +57,34 @@ impl Intrinsic for Json {
...
     let parse = Function::new(
         this.clone(),
         MutFn::new(move |cx: Ctx<'js>, args: Rest<Value<'js>>| {
-            call_json_parse(hold!(cx.clone(), args)).map_err(|e| to_js_error(cx, e))
+            call_json_parse(hold!(cx.clone(), args), default_parse.clone())
+                .map_err(|e| to_js_error(cx, e))
         }),
     )?;
 
+    // Explicitly set the function's name and length properties.
```
jeffcharles, line 73: Can we include the reason for explicitly setting the name and length properties in the comment?

crates/javy/src/apis/json.rs
```diff
@@ -74,46 +93,77 @@ fn register<'js>(this: Ctx<'js>) -> Result<()> {
...
             let mut string = val_to_string(this.clone(), args[0].clone())?;
             let bytes = unsafe { string.as_bytes_mut() };
-            json::parse(this, bytes)
+            json::parse(this.clone(), bytes).map_err(|original| {
+                if original.downcast_ref::<SError>().is_none() {
+                    return original;
+                }
+
+                // TODO: Use the actual result here.
+                let _e = original.downcast::<SError>().unwrap();
```
jeffcharles, line 126: Why are we doing this?

crates/javy/src/lib.rs
```diff
@@ -117,10 +122,28 @@ pub fn from_js_error(ctx: Ctx<'_>, e: JSError) -> Error {
...
+            // particularly for json::stringify. The process of transcoding will
+            // report the Serializer error, which is totally implementation
+            // dependent, in this case particular to serde_json::Error. To
+            // workaround this, we identify the exception via its string
+            // representation. This is not ideal, but its also fine as it only
+            // happens in the transcoding case.
+            //
+            // Ref: https://github.com/sfackler/serde-transcode/issues/8
+            if e.to_string()
+                .contains("JSError: Exception generated by QuickJS")
```
jeffcharles, line 137: Can we add an automated test case to cover this case?

crates/javy/src/serde/de.rs
```diff
@@ -78,6 +96,21 @@ impl<'js> Deserializer<'js> {
         }
         unreachable!()
     }
+
+    /// When stringifying, circular dependencies are not allowed. This function
```
jeffcharles, line 100: I might be missing something. Does stringifying use the deserialization path in serde? In my head it's a serialization operation.

crates/javy/src/serde/de.rs
```diff
@@ -119,32 +176,80 @@ impl<'de, 'a> de::Deserializer<'de> for &'a mut Deserializer<'de> {
...
-            let filter = Filter::new().enum_only().symbol().string();
+            ensure_supported(&self.value).and_then(|_| {
+                self.check_cycles()
+                    .map(|_| self.stack.push(self.value.clone()))
+            })?;
+
+            if let Some(f) = get_to_json(&self.value) {
+                let v: Value = f.call((This(self.value.clone()),))?;
+
+                // TODO: Must find a way to discard.
```
jeffcharles, line 216: Can you go into more detail?

jeffcharles (commented):
Do we need to take msgpack or other serde supported non-JSON formats into account with the changes we're making to the serde trait implementations?

crates/javy/src/lib.rs
saulecabrera: This one is extensively exercised in the stringify 262 test cases. As the comment suggests, it's an edge case of using transcode.

crates/javy/src/serde/de.rs
saulecabrera: It uses the deserializer, it's a bit counterintuitive, but we're deserializing from a JS value into any other format supported by serde.

crates/javy/src/serde/de.rs
saulecabrera: Oh this one is stale, I'll remove.

crates/javy-test-macros/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+    root: PathBuf,
+}
+
+impl Config262 {
+    fn validate(&self) -> anyhow::Result<()> {
+        let path: Box<Path> = self.root.clone().into();
+        if path.is_dir() {
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!("Invalid path"))
```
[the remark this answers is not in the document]
saulecabrera: I find this useful when we have `return Err` patterns, I'll see what can be updated.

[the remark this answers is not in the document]
saulecabrera: Oh yeah and also in the serde pieces, we can't use bail as the error type doesn't come from anyhow, instead the serde traits require a custom Error definition.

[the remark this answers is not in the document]
saulecabrera: If we'd like to enforce the usage of bail where possible, I wonder if there's a way to configure a custom clippy rule, that way we can delegate those stylistic lints to the tooling. Personally I don't have a strong opinion here.

crates/javy/src/apis/json.rs
saulecabrera: I forgot to use the actual error. For completeness, the spec tests don't care about the error message, they assert against the error type.

crates/javy-test-macros/src/lib.rs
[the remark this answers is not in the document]
saulecabrera: Added an update in the most obvious places.

saulecabrera:
> Do we need to take msgpack or other serde supported non-JSON formats into account with the changes we're making to the serde trait implementations?

Ah yeah, good question. To ensure that it's correct, we'd probably need to hook the spec test for msgpack, similar to what's done in this PR for JSON. I suspect that the in-the-wild usage for the msgpack ser/deserializer is close to none, so my preference would be to deprecate this feature instead. Also, these serializers have been in "experimental" state since I initially wrote them, until now, in which we're upgrading the status of JSON. I'll do the deprecation in a follow up though.

jeffcharles (commented):
Mostly looks good to me! Just noticed the error messages not including the type of error while seeing what the error message with the unbalanced double-quote looked like.

Re non-JSON formats, I agree that the msgpack transcoding probably isn't used widely. However, we were publicly exporting the `Serializer` and `Deserializer` from the `quickjs-wasm-rs` crate and someone could've been using those for any format (I also don't have any evidence they were in use) but it also looks like we're not exporting them from the Javy crate so I guess we're probably fine since they wouldn't be able to continue to use the serializer or deserializer anyway. I guess if someone complains, we can figure out options to support their use case at if that happens.

crates/javy/src/serde/err.rs
```diff
@@ -11,7 +12,13 @@ impl error::Error for Error {}
 impl fmt::Display for Error {
     fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Error::Custom(e) => formatter.write_str(&e.to_string()),
+            Error::Custom(e) => {
+                if let Some(e) = e.downcast_ref::<JSError>() {
+                    formatter.write_str(&format!("JSError: {}", &e.to_string()))
```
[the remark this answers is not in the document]
saulecabrera: The context is lost when triggering an error from the deserializer and the error is handled by the serializer (serde_json); it only happens in the case of stringify because the serde_json serializer is on the receiving end of the transcoding process. The introduction of `JSError: ....` intentional to be able to match against a known exception let the exception bubble up. See the serde transcode issue linked in https://github.com/bytecodealliance/javy/pull/662#discussion_r1633726753

crates/javy/src/apis/json.rs
```diff
@@ -74,46 +95,76 @@ fn register<'js>(this: Ctx<'js>) -> Result<()> {
...
 
-fn call_json_parse(args: Args<'_>) -> Result<Value> {
+fn call_json_parse<'a>(args: Args<'a>, default: Function<'a>) -> Result<Value<'a>> {
     let (this, args) = args.release();
 
     match args.len() {
-        0 => Err(anyhow!("Error: \"undefined\" is not valid JSON")),
+        0 => bail!(Exception::throw_syntax(
+            &this,
+            "undefined\" is not valid JSON"
```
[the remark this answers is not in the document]
saulecabrera: The type here is indeed a syntax error, because there are 262 tests that exercise this path and that assert the error type. I'm pretty sure that the lack of error type was not introduced here but instead was introduced when migrating to rquickjs. By investigating a bit more, the culprit seems to be here https://github.com/bytecodealliance/javy/blob/main/crates/core/src/execution.rs#L38, when calling `from_js_error`, which calls the `display` method on `JSError`. So to answer your question, all unhandled exceptions will lack types when displayed, because this behavior comes [from the `Display` implementation for `Exception` in rquickjs](https://docs.rs/rquickjs-core/0.6.2/src/rquickjs_core/value/exception.rs.html#276)

Even though this is not semantically changing the type of the error, I think it's probably worth fixing upstream, assuming that there's a way to inspect the exception type in QuickJS.

jeffcharles (approved):

PR #664, opened 2024-06-11
Introduce the `javy-config` crate

saulecabrera:
This commit prepares the terrain to bridge the gap in configurability between the CLI and Javy, the crate.
The intention behind introducing a new crate is to:

* Reduce code duplication and sync the options between the CLI and the core crate.
* Make it easy to pass the options into WebAssemlby by using bitflags.

This PR doesn't introduce any new functionality. A follow up PR will include new commands in the CLI which will make use of the share configuration.

## Checklist

### DIFF
--- a/crates/javy-config/src/lib.rs
+++ b/crates/javy-config/src/lib.rs
@@ -24,11 +24,25 @@
 use bitflags::bitflags;
 
 bitflags! {
+    #[derive(Eq, PartialEq)]
     pub struct Config: u32 {
         const OVERRIDE_JSON_PARSE_AND_STRINGIFY = 1;
         const JAVY_JSON = 1 << 1;
-        const JAVY_STREAM_IO = 1 << 1;
-        const REDIRECT_STDOUT_TO_STDERR = 1 << 2;
-        const TEXT_ENCODING = 1 << 3;
+        const JAVY_STREAM_IO = 1 << 2;
+        const REDIRECT_STDOUT_TO_STDERR = 1 << 3;
+        const TEXT_ENCODING = 1 << 4;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Config;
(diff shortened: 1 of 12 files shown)


saulecabrera:
In case the intention is not clear from the commit message/PR, the intention is to pass the config as a single u32 into WebAssembly, hence the usage of `bitflags`.

crates/javy-config/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+//!
+//! Currently, this crate includes only a subset of the available configuration
+//! options. The objective is to eventually encompass all configurable
+//! parameters found in [javy::Config].
+//!
+//! The selection of the current configuration options was influenced by the
+//! need to override non-standard defaults typically set during CLI invocations.
+//! These defaults often do not align with the preferences of the CLI users.
+//!
+//! In gneneral this crate should be treated as an internal detail and
```
jeffcharles, line 21:
```suggestion
//! In general this crate should be treated as an internal detail and
```

crates/javy-config/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+//! a contract between the CLI and the Javy crate.
+
+use bitflags::bitflags;
+
+bitflags! {
+    pub struct Config: u32 {
+        const OVERRIDE_JSON_PARSE_AND_STRINGIFY = 1;
+        const JAVY_JSON = 1 << 1;
+        const JAVY_STREAM_IO = 1 << 1;
```
jeffcharles, line 30: Should the last operand be incremented by 1 for this line and all of the following lines given we're using `1 << 1` for `JAVY_JSON`? Also is there anyway we could have tests or something else automated to catch if we're using the same value for two flags?

jeffcharles (approved):

PR #670, opened 2024-06-17
Statically link to the C runtime in MSVC

saulecabrera:
## Description of the change

Javy version 1.4.0 statically linked to the C runtime. Such version was built with Rust 1.75.

Javy version 2.0.0, built with Rust 1.79.0, which doesn't perform static linking by default. This commit adds a directive to preserve the same behavior as in v 1.4.0.

## Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,3 +4,9 @@
 # https://doc.rust-lang.org/cargo/reference/config.html
 [target.wasm32-wasi]
 rustflags = ["-C", "target-feature=+simd128"]
+
+# We want to ensure that all the MSVC dependencies are statically resolved and
+# included in the final CLI binary.
+# Ref: https://github.com/rust-lang/rust/pull/122268
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]


saulecabrera:
Testing

The best way to test this change at this time is to [download the Windows executable available in my fork](https://github.com/saulecabrera/javy/releases) and make sure that it runs correctly e.g. `javy --version`.

I considered adding a CI step, but unfortunately there are several considerations:

1. All Windows runners [are pre-provisioned with Visual C++](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md), and apparently: (i) there's no way to uninstall it (ii) there's no way to configure GitHub Actions to provide a fresh runner, without any dependencies. [Additionally the GitHub Actions team is not planning on supporting this feature.](https://github.com/actions/runner-images/issues/222). [The recommended way to modify Windows runners is through Chocolatey](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners), but as far as I investigated, it can't be used uninstall Visual C++ (not surprising).
2. The recommended approach, as per the link in the previous point is to introduce a custom runner. Which we could adopt if we wanted to, and if we do, my suggestion would be to do it a follow up PR. But generally speaking, the change in this PR are probably enough. If this issue were to happen again, there's a very easy fix that users can apply, which is installing Visual C++.

jeffcharles (approved):
Tested on my Windows PC without Visual Studio and it seems to work. I probably have a bunch of redistributable Visual Studio dlls installed but not really sure how to be more certain within a reasonable amount of time.

PR #678, opened 2024-06-29
Extract the `runner` module into its own crate

saulecabrera:
This commit extracts the utiility runner module, used in the integration tests into its own private crate.
The main motivation for this change is to enable sharing this functionality between integration tests and the upcoming `javy-fuzz` crate.

Additionally, this change promotes some dependencies to the workspace level, namely:

- `tempfile`
- `uuid`

## Checklist

### DIFF
--- a/crates/runner/src/lib.rs
+++ b/crates/runner/src/lib.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{bail, Result};
 use std::error::Error;
 use std::fmt::{self, Display, Formatter};
 use std::io::{self, Cursor, Write};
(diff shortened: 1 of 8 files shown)


crates/runner/src/lib.rs
```diff
@@ -10,10 +10,93 @@ use wasi_common::sync::WasiCtxBuilder;
...
+        self.input = path.into();
+        self
+    }
+
+    pub fn bin(&mut self, bin: impl Into<String>) -> &mut Self {
+        self.bin_path = bin.into();
+        self
+    }
+
+    pub fn wit(&mut self, wit: impl Into<PathBuf>) -> &mut Self {
```
jeffcharles, line 59: Arguably wit and world could be one function given they either both need to be provided or neither should be provided but this is fine

crates/runner/src/lib.rs
```diff
@@ -10,10 +10,93 @@ use wasi_common::sync::WasiCtxBuilder;
...
+    /// Root path. Used resolve the absolute path of the JS source.
+    root: PathBuf,
+    /// `javy` binary path.
+    bin_path: String,
+    /// The path to the wit file.
+    wit: Option<PathBuf>,
+    /// The name of the wit world.
+    world: Option<String>,
+    /// The logger capacity, in bytes.
+    capacity: usize,
```
jeffcharles, line 25: I think we can get rid of capacity. Initially I added it because I wanted to test a change for what should happen if the host-side buffer size is exceeded but IIRC we never altered anything in Javy so I think we're just always setting this to `usize::MAX`.

jeffcharles (approved):

PR #685, opened 2024-07-02
Add crypto hmac sha 256

ewalk153:
## Description of the change

Running crypto functions in Javascript quickly runs out of fuel, which rust is must more efficient at the job. This is a wrapper to expose the rust hmac-sha256 implementation to javy user code.

To conserve fuel when running crypto. As a bonus, the crypto function is not left to end users to get right.

## Checklist

### DIFF
--- a/crates/javy/src/config.rs
+++ b/crates/javy/src/config.rs
@@ -19,6 +19,7 @@ bitflags! {
         const OPERATORS = 1 << 12;
         const BIGNUM_EXTENSION = 1 << 13;
         const TEXT_ENCODING = 1 << 14;
+        const CRYPTO = 1 << 15;
     }
 }
 
@@ -38,7 +39,6 @@ bitflags! {
     pub(crate) struct JavyIntrinsics: u32 {
         const STREAM_IO = 1;
         const JSON = 1 << 1;
-        const CRYPTOX = 1 << 2;
     }
 }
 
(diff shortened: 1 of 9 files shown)


ewalk153:
This code still need a change log entry, on top of any polished suggested during code review.

crates/javy/src/config.rs
```diff
@@ -38,6 +38,7 @@ bitflags! {
     pub(crate) struct JavyIntrinsics: u32 {
         const STREAM_IO = 1;
         const JSON = 1 << 1;
+        const CRYPTOX = 1 << 2;
```
saulecabrera, line 41: Similar to above, given that this API is not namespaced through `Javy`, it might make sense to add it as part of `JSIntrinsics` instead, similar to `TEXT_ENCODING` (here https://github.com/bytecodealliance/javy/pull/685/files#diff-f9510ffe1a14a724c735bc8f7f3d0196470ccf82f09f399287be0db59d593cc6R21) -- that is, if we decide on start standardizing this API through Winter CG as discussed offline.

saulecabrera (commented):
Looks to be going in a good direction!

saulecabrera:
The cargo vet failures are expected, since you're introducing new dependencies. We can resolve those once we're getting close to merging this one.

ewalk153:
If someone else is interested in pushing this forward, feel free to reach out.

PR #690, opened 2024-07-02
Update `Cargo.lock`

saulecabrera:
The latest main contains an `out-of-sync` Cargo.lock, which is making the newly introduced `cargo metadata` invocation fail in CI


saulecabrera:
The reason behind the failure is that I added `publish = false` to the `test-macros` and `config` crates, and that needed to be reflected in the lock file.

jeffcharles (approved):

PR #696, opened 2024-07-04
Add Crypto HmacSha256 support

ewalk153:
## Description of the change

Conforming to the [WinterCG crypto Hmac convention](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/sign_verify/hmac.js), introduce HmacSha256 support. This can be extended to add other crypto functions in the future.

Uses promises, per the WinterCG convention.

Example code so far:
```
await crypto.subtle.sign({name: "HMAC", hash: "sha-256"}, "my secret and secure key", "input message");
```

And to be completely WinterCG complaint it will soon be:
```
(async (secret, body) => {
  let enc = new TextEncoder("utf-8");
  let algorithm = { name: "HMAC", hash: "SHA-256" };

  let key = await crypto.subtle.importKey("raw", enc.encode(secret), algorithm, false, ["sign", "verify"]);
  let signature = await crypto.subtle.sign(algorithm.name, key, enc.encode(body));
  
  return Array.from(new Uint8Array(signature)).map(byte => byte.toString(16).padStart(2, '0')).join('');
})("my secret and secure key", "input message");
```

TODO:
- [ ] add `crypto.subtle.importKey` and make this the required key parameter to `sign` - see [sign example code](https://gist.github.com/ewalk153/9127b47f02a6baf0aecb2012428daa76)
- [ ] change return value of `sign` to be a byte array that is converted to a string with user code such as: `let hexString = Array.from(new Uint8Array(signature)).map(byte => byte.toString(16).padStart(2, '0')).join('');`

There is a strong need for for function input authentication. HmacSha256 strikes a balance between instruction count and cryptographic safety as a sane starting point.

## Checklist


saulecabrera:
So apologies for the bikeshed here, the spec is a bit hard to navigagte, but after investigating a bit more, I believe that what we want is exposing [`crypto.suble.digest`](https://w3c.github.io/webcrypto/#SubtleCrypto-method-digest) rather than `createHmac` which is not standard yet (that's why it was not in the doc). 

Thankfully I think that this doesn't represent a huge change to the current state of your PR, in fact, I think it might make the implementation a bit easier. The only wrinkle to think about is that `crypto.subtle.digest` returns a promise, which I believe we can workaround by wrapping the result via: https://docs.rs/rquickjs/latest/rquickjs/struct.Promise.html#method.from_value, without having to actually defer the computation.

saulecabrera:
The other nice thing is that we can now refer to https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/digest/digest.https.any.js to compliance tests.

ewalk153:
Next step: change API signature to:
```
crypto.sutble.sign({name: "HMAC", hash: "sha-256"}, "my key", "my plaintext")
```

This aligns with the [wintercg spec](https://github.com/web-platform-tests/wpt/blob/master/WebCryptoAPI/sign_verify/hmac.js).

Per convo with @saulecabrera  and @jeffcharles

saulecabrera:
@ewalk153 I've pushed https://github.com/bytecodealliance/javy/pull/696/commits/fa3c661b24c546a6f6af69554eab29da5c8a8a69, see the commit description for details. Additionally, while working on these improvements, I found a bug with our event loop handling code (fixed in https://github.com/bytecodealliance/javy/pull/697), so I'd suggest rebasing once that one lands. I didn't fix all the unit tests in my last commit here, so there's a chance that we'd need to revisit those.

MurmeltierS:
sorry for chirping in: is this still being considered?

ewalk153:
Still looking to introduce a version of these crypto function into Javy, in some form.

ewalk153:
If someone else is interested in pushing this forward, feel free to reach out.

PR #697, opened 2024-07-08
Tidy up event loop handling code

saulecabrera:
This commit refactors the event loop handling code, mainly by extracting the common pieces into a reusable function.

Additionally, this commit ensures that the result of `Promise.finish` is corectly handled, which fixes execution of code with top level awaits.

Finally, this commit also ensures that we test the `experimental_event_loop` in CI for the CLI and core crates (follow-up to https://github.com/bytecodealliance/javy/pull/238)

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,14 +86,7 @@ jobs:
           path: target/wasm32-wasi/release/
 
       - name: Test CLI
-        env:
-          CARGO_PROFILE_RELEASE_LTO: off
-        run: |
-          cargo build --package=javy-core --target=wasm32-wasi --release
-          cargo test --package=javy-cli --release -- --nocapture
-
-          cargo build --package=javy-core --target=wasm32-wasi --release --features=experimental_event_loop
-          cargo test --package=javy-cli --features=experimental_event_loop --release -- --nocapture
+        run: CARGO_PROFILE_RELEASE_LTO=off cargo test --package=javy-cli --release -- --nocapture
 
       - name: Check benchmarks
         run: CARGO_PROFILE_RELEASE_LTO=off cargo check --package=javy-cli --release --benches
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,14 @@ jobs:
           path: target/wasm32-wasi/release/
 
       - name: Test CLI
-        run: CARGO_PROFILE_RELEASE_LTO=off cargo test --package=javy-cli --release -- --nocapture
+        env:
+          CARGO_PROFILE_RELEASE_LTO: off
+        run: |
+          cargo build --package=javy-core --target=wasm32-wasi --release
+          cargo test --package=javy-cli --release -- --nocapture
+
+          cargo build --package=javy-core --target=wasm32-wasi --release --features=experimental_event_loop
+          cargo test --package=javy-cli --features=experimental_event_loop --release -- --nocapture
 
       - name: Check benchmarks
         run: CARGO_PROFILE_RELEASE_LTO=off cargo check --package=javy-cli --release --benches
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -86,7 +86,14 @@ jobs:
...
       - name: Test CLI
-        run: CARGO_PROFILE_RELEASE_LTO=off cargo test --package=javy-cli --release -- --nocapture
+        env:
+          CARGO_PROFILE_RELEASE_LTO: off
+        run: |
+          cargo build --package=javy-core --target=wasm32-wasi --release
+          cargo test --package=javy-cli --release -- --nocapture
+
+          cargo build --package=javy-core --target=wasm32-wasi --release --features=experimental_event_loop
+          cargo test --package=javy-cli --features=experimental_event_loop --release -- --nocapture
```
jeffcharles, line 96: This will result in CI taking materially longer to complete. Can we do the recompilation of `javy-core` and `javy-cli` with the `experimental_event_loop` feature as a separate GitHub Actions workflow? Similar to what we do for WPT. Or maybe rename the WPT pipeline and do a `cargo test` with the feature enabled and also perform a WPT run in that workflow? That way we continue to compile the `javy-core` module with the additional feature in parallel instead of serially.

crates/cli/tests/integration_test.rs
```diff
@@ -96,7 +96,11 @@ fn test_logging() -> Result<()> {
         "hello world from console.log\nhello world from console.error\n",
         logs.as_str(),
     );
-    assert_fuel_consumed_within_threshold(34169, fuel_consumed);
+    if cfg!(feature = "experimental_event_loop") {
+        assert_fuel_consumed_within_threshold(39602, fuel_consumed);
```
jeffcharles, line 100: I won't hold up merging on this but I wonder if it's worth moving the cfg check into `assert_fuel_consumed_within_threshold` and only checking the fuel if the feature is off. Then we only need to track one number for every test.

saulecabrera:
There's something strange about this failure, it's not failing locally. Let me look into it. I'll report back once it's solved.

.github/workflows/ci.yml
saulecabrera: Oh right, yeah good point. I wouldn't use wpt, but a separate workflow seems reasonable.

crates/cli/tests/integration_test.rs
saulecabrera: Considered it. However, it's important to keep track of fuel used regardless of the feature, especially this one, as it'll inform what's the impact when enabling it by default. It's possible that I'll need to revisit/change  this though, see: https://github.com/bytecodealliance/javy/pull/697#issuecomment-2214136414

saulecabrera: Reverted this change, there was no need for it.

.github/workflows/ci.yml
saulecabrera: Fixed.

.github/workflows/cli-features.yml
```diff
@@ -0,0 +1,26 @@
+# Tests extra CLI features as and their dependency with core features.
```
jeffcharles, line 1: I think there's a typo in this comment around _features as and their dependency_

.github/workflows/cli-features.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          os: linux
+
+      - name: Test `experimental_event_loop`
+        run: |
+          cargo build --package=javy-core --target=wasm32-wasi --release --features=experimental_event_loop
+          cargo test --package=javy-cli --features=experimental_event_loop --release -- --nocapture
+
+
+
```
jeffcharles, line 26:
```suggestion
```

jeffcharles (approved):

PR #699, opened 2024-07-08
Initial skeleton for fuzzing infrastructure

saulecabrera:
This set of commits introduce the initial skeleton for Javy's fuzzing
infrastructure.

This commit focuses on the `json-differential` target, to check Javy's
SIMD-based, custom `JSON.parse` and `JSON.stringify` implementations.

The implementation aims to take full advantage of libfuzzer's coverage
capabilities and that's the reason why it uses the native version of
QuickJS instead of the Wasm version.

## Checklist


Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
   "crates/test-macros",
   "crates/config",
   "crates/runner",
+  "fuzz",
```
saulecabrera, line 12: This is intentional. Added it outside of the `crates` directory, given the infrastructure nature of this crate.

.github/workflows/check-fuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  check-fuzz:
+    name: Build fuzz targets
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install nightly
+
+        run: rustup install nightly-2024-07-07
```
saulecabrera, line 20: Pinning to the most recent nightly version, and the one that I used locally for testing.

.github/workflows/check-fuzz.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Install cargo fuzz
+        run: cargo install cargo-fuzz --version=0.12.0
```
saulecabrera, line 22: Same here.

saulecabrera:
Decided to drop clippy lints for fuzz targets in https://github.com/bytecodealliance/javy/pull/699/commits/7d94ba8179c80ef68b6b987ca1c1420aecd5a2e5. Originally I wanted to use clippy + nightly to ensure that fuzz targets are linted correctly, but I was not able to figure out a clean way to have clippy only lint the package and not its dependencies. I don't want to spend too much cycles on this, we can revisit if we think that linting the fuzz targets is crucial. See https://github.com/rust-lang/rust-clippy/issues/3837#issuecomment-617413116 for more details.

.github/workflows/check-fuzz.yml
```diff
@@ -0,0 +1,42 @@
...
+  check-fuzz:
+    name: Build fuzz targets
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Read pinned nightly version
+        id: nightly_version
+        shell: bash
+        run: |
+          NIGHTLY_VERSION=$(cat pinned-nightly-version)
```
jeffcharles, line 22: What was the reason for putting this and the cargo fuzz version in a file instead of using a top-level `env` for both? I don't see them referenced outside this file.

jeffcharles (approved):
Just have the one question but it's pretty minor

.github/workflows/check-fuzz.yml
saulecabrera: The reason is very subjective: I feel that's easier to update the version when it's defined at the root than having to modify the action itself.

PR #700, opened 2024-07-11
Improve cycle checks in  `JSON.stringify`

saulecabrera:
Prior to this commit, the following program will erroneously fail, with a `circular dependency` error:

    const template = {
      foo: "foo",
      bar: "bar",
    };
    const x = [];

    x.push(template);
    x.push(template);
    x.push(template);
    x.push(template);
    x.push(template);
    x.push(template);

    console.log(JSON.stringify(x));

The logic for testing for circular dependencies didn't correctly manage the stack state and therefore pushing the same object multiple times to an array resulted in a false-positive cycle.

Cycles can only really happen in Arrays and Objects. This commit solidifies the approach to checking for cycles, by scoping cycle checks and cycle stack management to the `MapAccess` and `SeqAccess` structs.

## Checklist

### DIFF
--- a/crates/javy/src/serde/de.rs
+++ b/crates/javy/src/serde/de.rs
@@ -103,7 +103,7 @@ impl<'js> Deserializer<'js> {
         let v = self
             .stack
             .pop()
-            .ok_or_else(|| anyhow!("No entries found in the deserialiazer stack"))?;
+            .ok_or_else(|| anyhow!("No entries found in the deserializer stack"))?;
         Ok(v)
     }
 
@@ -442,7 +442,7 @@ impl<'a, 'de> de::SeqAccess<'de> for SeqAccess<'a, 'de> {
                 self.de.value = Null.into_value(self.seq.ctx().clone())
             }
             self.index += 1;
-            // Check cycles right before kicking the deserialiazation for the
+            // Check cycles right before starting the deserialization for the
             // sequence elements.
             self.de.check_cycles()?;
             seed.deserialize(&mut *self.de).map(Some)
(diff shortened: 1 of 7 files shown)


crates/javy/src/serde/de.rs
```diff
@@ -97,6 +98,15 @@ impl<'js> Deserializer<'js> {
         unreachable!()
     }
 
+    /// Pops the last visited value present in the stack.
+    fn pop_visited(&mut self) -> Result<Value<'js>> {
+        let v = self
+            .stack
+            .pop()
+            .ok_or_else(|| anyhow!("No entries found in the deserialiazer stack"))?;
```
jeffcharles, line 106: Nit
```suggestion
            .ok_or_else(|| anyhow!("No entries found in the deserializer stack"))?;
```

crates/javy/src/serde/de.rs
```diff
@@ -385,8 +442,13 @@ impl<'a, 'de> de::SeqAccess<'de> for SeqAccess<'a, 'de> {
                 self.de.value = Null.into_value(self.seq.ctx().clone())
             }
             self.index += 1;
+            // Check cycles right before kicking the deserialiazation for the
+            // sequence elements.
```
jeffcharles, line 446: Can you clarify this comment a little? I'm not sure what _kicking_ is referring to in this case. Also `deserialiazation` -> `deserialization`.

jeffcharles (approved):

crates/javy/src/serde/de.rs
saulecabrera: Not too much to clarify, other than the comment was half-baked and not ready. Fixed.

PR #701, opened 2024-07-11
CLI: Allow Input from stdin and Output to stdout

konradkoschel:
## Description of the change

In this PR, the CLI crate is enhanced so that a hyphen (`-`) can be passed for the input argument or the output option instead of a file name indicating that stdio streams should be used.

### Possible usage

```sh
# Use stdin for input and file for output
echo "console.log('Hello World!');" | javy compile - -o my.wasm

# Pipe wasm output into another command
javy compile my.js -o - | wasm-opt ...

# Use no files at all
echo "console.log('Hello World!');" | javy compile - -o - | wasm-opt ...
```

I built a wrapper around Javy. The wrapper is not inherently working with files, but to make it work with Javy, I need to write the JavaScript to the filesystem and read the WASM from the filesystem. 

Generally, the changes make the CLI of Javy more flexible.

## Checklist

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -12,7 +12,7 @@ use clap::Parser;
 use js::JS;
 use std::fs;
 use std::fs::File;
-use std::io::Write;
+use std::io::{Read, Write};
 use wasm_generator::dynamic as dynamic_generator;
 
 fn main() -> Result<()> {
@@ -21,7 +21,14 @@ fn main() -> Result<()> {
     match args.command {
         Command::EmitProvider(opts) => emit_provider(&opts),
         Command::Compile(opts) => {
-            let js = JS::from_file(&opts.input)?;
+            let js = match opts.input.to_str() {
+                Some("-") => {
+                    let mut content = String::new();
+                    std::io::stdin().read_to_string(&mut content)?;
+                    JS::from_string(content)
+                }
+                _ => JS::from_file(&opts.input)?,
+            };
             let exports = match (&opts.wit, &opts.wit_world) {
                 (None, None) => Ok(vec![]),
                 (None, Some(_)) => Ok(vec![]),
@@ -33,7 +40,12 @@ fn main() -> Result<()> {
             } else {
                 static_generator::generate(&js, exports, opts.no_source_compression)?
             };
-            fs::write(&opts.output, wasm)?;
+            match opts.output.to_str() {
+                Some("-") => {
+                    std::io::stdout().write_all(&wasm)?;
+                }
+                _ => fs::write(&opts.output, wasm)?,
+            }
             Ok(())
         }
     }
--- a/crates/cli/src/js.rs
+++ b/crates/cli/src/js.rs
@@ -32,7 +32,7 @@ pub struct JS {
 }
 
 impl JS {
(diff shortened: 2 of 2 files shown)


saulecabrera (commented):
Thanks for the PR -- generally these changes seem reasonable to me. One thing to note is that we're currently thinking of revamping Javy's CLI interface (see https://github.com/bytecodealliance/javy/issues/702).

Given that our intention is to move in the direction described in the issue linked above, it probably makes sense to include this change as part of the new `build` command. That said, you're not expected to take on the CLI changes, for this change to land. However, I'd like to ask if you're ok waiting until we at least introduce the command that will replace the current `compile` command in order to merge this?

crates/cli/src/main.rs
```diff
@@ -21,7 +21,14 @@ fn main() -> Result<()> {
     match args.command {
         Command::EmitProvider(opts) => emit_provider(&opts),
         Command::Compile(opts) => {
-            let js = JS::from_file(&opts.input)?;
+            let js = match opts.input.to_str() {
+                Some("-") => {
```
saulecabrera, line 25: On the technical side of this change, any particular reason why you decided chose hyphens over:

- Declaring the `input` and `output` options as optional via `Option<T>`
- Default to `stdin` / `stdout` accordingly if either or both are not specified

konradkoschel:
> Thanks for the PR -- generally these changes seem reasonable to me. One thing to note is that we're currently thinking of revamping Javy's CLI interface (see #702).
> 
> Given that our intention is to move in the direction described in the issue linked above, it probably makes sense to include this change as part of the new `build` command. That said, you're not expected to take on the CLI changes, for this change to land. However, I'd like to ask if you're ok waiting until we at least introduce the command that will replace the current `compile` command in order to merge this?

Hi, thanks for the fast response! I agree on the point that it makes sense to incorporate the functionality into the upcoming `build` command.
From my point of view, the option to use the standard streams as input and output for Javy is a nice addition, but it is not too urgent.
So I can surely wait :)

crates/cli/src/main.rs
konradkoschel: For stdin input, I would argue that it is common practice to use a hyphen as the argument to indicate this explicitely. For the output, I didn't want to break the current default behavior (i.e. using `index.wasm`).

I see the point though, that especially the combination of both (`-o - -` / `- -o -`) may look a bit weird. In the end, I wanted to primarily make a proposal for the functionality – the implementation details can be discussed.

As this will anyway be implemented as part of the new `build` command, I would default the output to stdout as suggested by you.

saulecabrera:
Hi @konradkoschel, wanted to let you know that most of the backing infrastructure for the CLI redesign has landed on main, and I think it's now a good time to rebase, update and move forward with this change. Let me know if you have any questions.

konradkoschel:
> Hi @konradkoschel, wanted to let you know that most of the backing infrastructure for the CLI redesign has landed on main, and I think it's now a good time to rebase, update and move forward with this change. Let me know if you have any questions.

Hi Saul,

I updated the branch. Could you check if the code changes are fine?

As discussed previously, I made the binary output the default instead of `-o -` as in the original implementation. Tests run successful on my side

PR #712, opened 2024-08-02
Fix CI

jeffcharles:
## Description of the change

Removing deprecated crates from the workspace, the fmt checks for deprecated crates from the makefile, and a few updates to appease the new version of Clippy:
- changing `&mut store` to `store.as_mut_context()`
- adjust indentation in one set of docs
- mark two imports that are only used if a particular feature is enabled are only imported when that feature is enabled

CI always uses the latest stable version of Clippy and that version of Clippy introduced new checks which our code fails. I adjusted the workspace and the `Makefile` since I don't want to make changes to the deprecated `quickjs-wasm-sys` or `quickjs-wasm-rs` crates but still want to be able to run `make fmt` to see all failing Clippy checks that would show up in CI.

For the `&mut store` to `store.as_mut_context()` changes, Clippy seems to sometimes get confused when passing `&mut store` to a Wasmtime method that takes `store: impl AsContextMut` and states we should only pass `store`. Unfortunately the Rust compiler obviously complains when passing `store` without the `&mut` if `store` is referenced again later because Rust will treat it as a move. At first I wrote I just updated call sites where Clippy complained but then it looks inconsistent within a function or file where sometimes we're passing `store.as_context_mut()` and sometimes we're passing `&mut store` and it's not obvious why we're using the first in some places and the second in others. So, I decided to make passing stores consistent since the behaviour is identical anyway.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,16 +587,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "core-foundation"
-version = "0.9.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
-dependencies = [
- "core-foundation-sys",
- "libc",
-]
-
-[[package]]
 name = "core-foundation-sys"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


saulecabrera (approved):
The change looks good to me. I'm concerned with the clippy/rustc dance, though. I'm wondering, as I've mentioned informally in the past, if we should consider removing clippy altogether or at the very least reduce the lints, similar to https://github.com/bytecodealliance/wasmtime/blob/main/Cargo.toml#L174, to avoid this from happening in the future. Personally I don't find it useful and can't remember an instance in which it was useful to the project.

Issue #702, opened 2024-07-11
Restructure the CLI


saulecabrera:
This proposal suggests a restructuring of Javy's CLI interface to:

- Enhance flexibility in specifying code generation options and JavaScript engine features.
- Provide semantically meaningful command and sub-command names where applicable.

#### Command Flexibility

Javy currently offers a [wide range of configuration options](https://github.com/bytecodealliance/javy/blob/main/crates/javy/src/config.rs) that control the JavaScript language features available to the generated code. However, there are limitations:

- There is currently no way to specify these options directly from the CLI.
- Other code-generation options, such as [name section generation](https://github.com/bytecodealliance/javy/blob/de8431ff25868f96a6338083748f691fac46bba0/crates/cli/src/wasm_generator/transform.rs#L39) and [optimization options](https://github.com/bytecodealliance/javy/blob/main/crates/cli/src/wasm_generator/static.rs#L139), cannot be specified through the CLI. These options are crucial for profiling Javy-generated modules using tools like [wasmprof](https://github.com/Shopify/wasmprof). Currently, users must manually update these options, which is not ideal.
- The current default settings are not well-suited to all use cases, as highlighted in [issue #628](https://github.com/bytecodealliance/javy/issues/628).

#### Semantically Meaningful Commands, Sub-Commands and Options
The CLI currently features two main commands: `compile` and `emit-provider`. The compile command does not accurately reflect its actions and could be better named to represent actual AOT compilation of JavaScript, a feature we plan to explore in the future.

#### Proposal

The high-level proposal includes:

- Deprecating the `compile` command in favor of a new command, tentatively named `build`. Avoid introducing breaking changes or new features through the `compile` command.
- Enhance the `build` command with meaningful options to control JavaScript and code generation features as described above.
- Updating the `compile` command to emit a deprecation notice, indicating that it will be deprecated in the next major release of the CLI.
  - The ideal timeline for transitioning from issuing a deprecation notice to actual deprecation is not clear. I propose to discuss this issue on our Zulip stream and make a decision based on the feedback received.
  
  
#### CLI Interface

The proposal suggests, according to the bullet points above, to have the new `build` command look like:

```sh
$ javy build --help

Builds a Wasm module from a JS source
Usage: javy build <OPTIONS> <JS>

Arguments:
  <JS> The JavaScript source to be used as input for building the WebAssembly module.

Options:

-J, --js <KEY[=VAL[,..]]> JavaScript runtime options.
[...]

PR #721, opened 2024-08-13
Introduce the `build` command in the CLI

saulecabrera:
This commit introduces a `build` command in the CLI, which behaves exactly as the `compile` command and introduces a deprecation warning to the existing `compile` command.

Additionally, as a preparation for the CLI redesign discussed in https://github.com/bytecodealliance/javy/issues/702, this commit introduces a small refactoring of the code generation process, by:

* Introducing a code generator builder, to abstract and validate all the code generation option combinations.
* Introducing proper static and dynamic code generator modules, making it easier to divide the responsibilities of each.

The main motivation for the refactoring is to:

  * Make it easier to finalize the CLI redesign.
  * Share code between the `compile` and `build` command while they must be equivalent.
  * Make it easier to keep the `compile` command frozen while it becomes deprecated and at the same time evolve the `build` command independently.

NB: Given that the `compile` and `build` command are exactly the same, and that this change is purely mechanical; this change doesn't introduce integration tests for the `build` command. The plan is to introduce tests once other options are added to the `build` command as part of the CLI redesign. Alternatively, tests can be added, however, it would require either duplicating the entire test suite or adding testing infrastructure to minimize duplication, which will increase the size of the change.

## Checklist


crates/cli/src/main.rs
```diff
@@ -1,38 +1,57 @@
...
     let args = Cli::parse();
 
-    match args.command {
-        Command::EmitProvider(opts) => emit_provider(&opts),
-        Command::Compile(opts) => {
+    match &args.command {
+        Command::EmitProvider(opts) => emit_provider(opts),
+        c @ Command::Compile(opts) | c @ Command::Build(opts) => {
+            if c.is_compile() {
+                println!(
```
jeffcharles, line 24: I think this should be emitted to stderr using `eprintln!` instead of stdout similar to how `--help` works.

jeffcharles (commented):
Looks good!

I think we should also update the help text for `compile` to indicate it's deprecated and update the README as well.

jeffcharles (approved):
🚀

PR #734, opened 2024-08-22
Add `-C` flag to build command

jeffcharles:
## Description of the change

I want to get early feedback on adding support for the `-C` flag documented in #702 so I can iterate on the draft as we figure out what makes sense to change.

There's a few changes:
- `CompileAndBuildCommandOpts` -> `CompileCommandOpts` and `BuildCommandOpts` including splitting out the handling of the `compile` and build` subcommands.
- `BuildCommandOpts` introduces a few extra types to try and get the repeated `-C` flags to play nicely with `clap`. There's a lot of copying triple slash comments around, none of it is picked up by `clap` so not sure if there's something better we can do. I looked at implementing `ValueEnum` as well but it looks like that's intended _only_ for enums with entirely unit variants. There's also a manual step to translate the `Vec<CodegenOption>` to `CodegenOptionGroup` since `clap` out-of-the-box only allows specifying the `-C` flag multiple times if the type associated with the option is a `Vec`.
- Changes to `javy-runner` to default to using the `build` command and support the arguments `build` supports unless a method has been invoked on the builder to use `compile` instead. I'm open to a different API.
- Adding `run_with_compile_and_build` to most tests to check that the build and compile commands function correctly. This could make more sense as a macro, I thought I'd start with a function because IMO it's easier to follow what the code is doing.

#702

## Checklist

### DIFF
--- a/crates/cli/src/commands.rs
+++ b/crates/cli/src/commands.rs
@@ -90,7 +90,7 @@ pub struct BuildCommandOpts {
 -C dynamic[=y|n] -- Creates a smaller module that requires a dynamically linked QuickJS provider Wasm module to execute (see `emit-provider` command).
 -C wit=path -- Optional path to WIT file describing exported functions. Only supports function exports with no arguments and no return values.
 -C wit-world=val -- Optional WIT world name for WIT file. Must be specified if WIT is file path is specified.
--C no-source-compression[=y|n] -- Disable source code compression, which reduces compile time at the expense of generating larger WebAssembly files.
+-C source-compression[=y|n] -- Enable source code compression, which generates smaller WebAssembly files at the cost of increased compile time. Defaults to enabled.
     "
     )]
     /// Codegen options.
@@ -115,8 +115,8 @@ pub struct CodegenOptionGroup {
     /// Optional path to WIT file describing exported functions.
     /// Only supports function exports with no arguments and no return values.
     pub wit_world: Option<String>,
-    /// Disable source code compression, which reduces compile time at the expense of generating larger WebAssembly files.
-    pub no_source_compression: bool,
+    /// Enable source code compression, which generates smaller WebAssembly files at the cost of increased compile time. Defaults to enabled.
+    pub source_compression: bool,
 }
 
 #[derive(Clone, Debug)]
@@ -130,8 +130,8 @@ pub enum CodegenOption {
     /// Optional path to WIT file describing exported functions.
     /// Only supports function exports with no arguments and no return values.
     WitWorld(String),
-    /// Disable source code compression, which reduces compile time at the expense of generating larger WebAssembly files.
-    NoSourceCompression(bool),
+    /// Enable source code compression, which generates smaller WebAssembly files at the cost of increased compile time. Defaults to enabled.
+    SourceCompression(bool),
 }
 
 impl FromStr for CodegenOption {
@@ -156,11 +156,11 @@ impl FromStr for CodegenOption {
                     .ok_or_else(|| anyhow!("Must provide value for wit-world"))?
                     .to_string(),
             ),
-            "no-source-compression" => Self::NoSourceCompression(match value {
+            "source-compression" => Self::SourceCompression(match value {
                 None => true,
                 Some("y") => true,
                 Some("n") => false,
-                _ => bail!("Invalid value for no-source-compression"),
+                _ => bail!("Invalid value for source-compression"),
             }),
             _ => bail!("Invalid codegen key"),
         };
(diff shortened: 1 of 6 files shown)


crates/cli/src/commands.rs
```diff
@@ -73,9 +74,121 @@ pub struct CompileAndBuildCommandOpts {
...
+    /// module to execute (see `emit-provider` command).
+    Dynamic(bool),
+    /// Optional path to WIT file describing exported functions.
+    /// Only supports function exports with no arguments and no return values.
+    Wit(PathBuf),
+    /// Optional path to WIT file describing exported functions.
+    /// Only supports function exports with no arguments and no return values.
+    WitWorld(String),
+    /// Disable source code compression, which reduces compile time at the expense of generating larger WebAssembly files.
+    NoSourceCompression(bool),
```
saulecabrera, line 134: I had thought of renaming this to `SourceCompression(bool)` and making it enabled by default. Mostly because I feel that it becomes easier to reason about as a boolean flag.

crates/cli/src/commands.rs
```diff
@@ -73,9 +74,121 @@ pub struct CompileAndBuildCommandOpts {
...
+pub enum CodegenOption {
+    /// Creates a smaller module that requires a dynamically linked QuickJS provider Wasm
+    /// module to execute (see `emit-provider` command).
+    Dynamic(bool),
+    /// Optional path to WIT file describing exported functions.
+    /// Only supports function exports with no arguments and no return values.
+    Wit(PathBuf),
+    /// Optional path to WIT file describing exported functions.
+    /// Only supports function exports with no arguments and no return values.
+    WitWorld(String),
```
saulecabrera, line 132: What do you think of using `crate::codegen::WitOptions`here?

crates/runner/src/lib.rs
```diff
@@ -27,6 +28,8 @@ pub struct Builder {
     built: bool,
     /// Preload the module at path, using the given instance name.
     preload: Option<(String, PathBuf)>,
+    /// Whether to use the `compile` or `build` command.
+    use_compile: bool,
```
saulecabrera, line 32: Could we make this an enum instead? (i.e., `command: JavyCommand`) I understand that this is probably temporary, until the `compile` command is fully deprecated, however it'll probably be easier to extend/refactor later by chasing down an enum rather than a boolean.

crates/cli/src/commands.rs
```diff
@@ -73,9 +74,121 @@ pub struct CompileAndBuildCommandOpts {
...
+    /// Disable source code compression, which reduces compile time at the expense of generating larger WebAssembly files.
+    NoSourceCompression(bool),
+}
+
+impl FromStr for CodegenOption {
```
saulecabrera, line 137: We had briefly discussed using https://docs.rs/clap/latest/clap/builder/trait.ValueParserFactory.html -- although the current approach seems to work well enough for the current PR. We can always iterate and refactor later.

saulecabrera (commented):
The direction looks good to me! Left some thoughts and minor comments.

crates/cli/src/main.rs
```diff
@@ -52,6 +51,29 @@ fn main() -> Result<()> {
...
+            let js = JS::from_file(&opts.input)?;
+            let codegen: CodegenOptionGroup = opts.codegen.clone().into();
+            let mut builder = CodeGenBuilder::new();
+            builder
+                .wit_opts(WitOptions::from_tuple((
+                    codegen.wit.clone(),
+                    codegen.wit_world.clone(),
+                ))?)
+                .source_compression(!codegen.no_source_compression)
+                .provider_version("2");
```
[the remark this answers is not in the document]
jeffcharles: I figured we'd bump this in a future PR (probably the one with where I add the `-J` flag) since we're not making any changes to the provider in this PR and there's already a lot going on.

crates/cli/src/commands.rs
jeffcharles: I should've mentioned in the description, I looked briefly at using it. My understanding is it works well if you need to keep track additional state outside of parsing the string to whatever type you're parsing it too. But the tradeoff is it forces you to define an additional type for the parser for the value, which is one more type to define. Ultimately since we have to collect a vector of _something_, I figured it would be easier to just write the state we need to collect to that _something_. That said, if we want to ensure there are no duplicate keys present and error out in the option collection phase, then having a custom parser would make that possible. Right now if an option is specified more than once, it'll take the last value as the one to use.

crates/cli/src/commands.rs
jeffcharles: I think it would make more sense to do that in the `CodegenOptionGroup` and update the `From<Vec<CodegenOption>>` implementation. The `wit` and `wit-world` arguments are parsed with two separate calls to the parser. So you can either store the intermediate values in this `Vec<CodegenOption>` (which is what I'm doing) and merge later or store the intermediate `wit` and `wit-world` values in the parser as state and then figure out some way to get that state out of the parser outside of clap (since clap_derive doesn't provide a great way to get the state out automatically).

saulecabrera: Yeah, you're right. I realized I placed this comment in the wrong place.

crates/cli/src/commands.rs
saulecabrera: >  That said, if we want to ensure there are no duplicate keys present and error out in the option collection phase, then having a custom parser would make that possible

In the long run, I think we'd want to ensure this. 

A custom parser also gives you more control over how to define/manage your group options and group help (e.g., you can have the help be automatically generated from the options without having to manually update the help string manually every time, which might be error prone).

I think it's fine to leave it as it for this first iteration, we can always refactor later.

saulecabrera (approved):

PR #739, opened 2024-08-27
Add JS runtime option infra and redirect-stdout-to-stderr option

jeffcharles:
## Description of the change

The only functional change in this PR is adding support for `-J redirect-stdout-to-stderr[=y|n]` with a default of `y` to `javy build`. There are no other functional changes and behaviour without specifying that flag doesn't change since the existing code uses `y` for that setting implicitly.

At the code level, there's a few related changes in this PR:
1. Changing the static and dynamic codegen builders to use dedicated `build` functions instead of a single shared function so I can pass an argument to the static codegen build function and not to the dynamic build function. This involved deleting some code that is no longer referenced.
2. Adding a `JsRuntimeOptionGroup` struct, a `JsRuntimeOption` enum, and trait implementations to convert a `Vec<JsRuntimeOption>` to a `JsRuntimeOptionGroup`, and from a `JsRuntimeOptionGroup` to a `javy_config::Config` so we have a way of representing command line options for JS runtime settings.
3. For static module generation, use the `JsRuntimeOptionGroup` struct to generate the `javy_config::Config` instead of using `Config::all()` and pass it into the codegen struct.
4. For dynamic module generation, return an error if the JS runtime settings are different than the default settings.
5. Implement `Default` for the `CodegenOptionGroup` and use the defaults for converting from `Vec<CodegenOption>` to `CodegenOptionGroup` to keep it consistent with how `JsRuntimeOptionGroup` works.
6. Update the testing infrastructure to work with the new argument.
7. Update the testing infrastructure so we can inspect stderr from Javy when a compile or build command fails.

See #702. I'm implementing this similarly to how I did for the codegen options. I'm also opting to return an error when generating a dynamic module if any of the JS runtime options are set to non-default values since those runtime options won't be applied to the generated module (the provider defines the runtime options). The refactoring of the codegen builder was because I wanted to try to make invalid state non-representable in the non-test code.

I can probably break this PR up if it's too large. I also tried avoiding any functional changes that I didn't have to so the PR could be kept smaller. Hence not bumping the provider version yet.

## Checklist

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -6,9 +6,9 @@ mod wit;
 
 use crate::codegen::WitOptions;
 use crate::commands::{Cli, Command, EmitProviderCommandOpts};
-use anyhow::{bail, Result};
+use anyhow::Result;
 use clap::Parser;
-use codegen::CodeGenBuilder;
+use codegen::{CodeGenBuilder, DynamicGenerator, StaticGenerator};
 use commands::{CodegenOptionGroup, JsRuntimeOptionGroup};
 use javy_config::Config;
 use js::JS;
@@ -44,11 +44,11 @@ fn main() -> Result<()> {
                 .source_compression(!opts.no_source_compression)
                 .provider_version("2");
 
+            let config = Config::all();
             let mut gen = if opts.dynamic {
-                builder.build_dynamic()?
+                builder.build::<DynamicGenerator>(config)?
             } else {
-                let config = Config::all();
-                builder.build_static(config)?
+                builder.build::<StaticGenerator>(config)?
             };
 
             let wasm = gen.generate(&js)?;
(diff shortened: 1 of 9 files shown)


jeffcharles:
The existing fuel consumption number on main is 37,309 for the log test is when compiled run on my Macbook.

crates/cli/src/main.rs
```diff
@@ -63,10 +65,14 @@ fn main() -> Result<()> {
                 .source_compression(codegen.source_compression)
                 .provider_version("2");
 
+            let js_runtime_options: JsRuntimeOptionGroup = opts.js_runtime.clone().into();
             let mut gen = if codegen.dynamic {
-                builder.build::<DynamicGenerator>()?
+                if js_runtime_options != JsRuntimeOptionGroup::default() {
+                    bail!("Cannot set JS runtime options when building a dynamic module");
+                }
+                builder.build_dynamic()?
```
saulecabrera, line 73: On a second thought, I wonder if the builder is better suited to perform this validation? 

That way:

1. We can keep the current `build` method, shared across dynamic/static codegen.
2. Perform the validation of the option groups through `javy::Config`
3. Keep main as slim as possible i.e., avoid baking in builder specific logic.

saulecabrera (commented):
Looks great to me. Left one comment that I think makes sense to address before landing this one!

crates/cli/src/main.rs
saulecabrera: In other words, we could modify the existing build method signature to be:

`fn build<T>(config: javy_config::Config) -> Result<..>`

jeffcharles: At some point in the future, if the defaults for `build` and `compile` diverge, the logic for validating the config for dynamic is going to depend on whether the build or compile command was used. But we can deal with that case when it actually comes up I guess.

saulecabrera (approved):

PR #740, opened 2024-08-29
Clarify ES2023 support

sandstrom:
Since it's now based on 2024-01-13 of QuickJS.

https://bellard.org/quickjs/

The exact version of QuickJS isn't specified somewhere. Don't know if that's something we want to spell out or not? I left it out for now.


saulecabrera (approved):

saulecabrera:
Thanks!

PR #748, opened 2024-09-06
Add support for `-J` option group

saulecabrera:
This commit adds support for the `-J` option group and enables `javy-json` and `override-json-parse-and-stringify` by default.

Additionally, this change improves how the javy cli integration tests are defined by introducing a `javy_cli_test` which:

* Produces a test per supported command (`compile` and `build`)
* Provides a set of defaults depending on the command under test

Generating a test per command improves the developer experience of determining failures on particular command configurations.

Lastly, this change bumps the provider to v3, given that new funcionality is introduced as part of enabling `javy-json` and `override-json-parse-and-stringify`. In order to test that the `compile` command remains "frozen", using version 2 of the provider, this commit introduces a static copy of the v2 provided downloaded from the releases section.


saulecabrera:
The failures in CI are caused by https://github.com/rust-lang/rust/issues/123748; in CI, we're using the latest stable, which got updated today. I can repro those failures locally by updating the toolchain.

crates/cli/src/commands.rs
```diff
@@ -228,6 +237,17 @@ option_group! {
...
         /// Whether to redirect the output of console.log to standard error.
         RedirectStdoutToStderr(bool),
+        /// Whether to enable the `Javy.JSON` builtins.
+        JavyJson(bool),
+        /// Whether to enable the `Javy.readSync` and `Javy.writeSync` builtins.
+        JavyStreamIo(bool),
+        /// Whether to override the `JSON.parse` and `JSON.stringify`
+        /// implementations with an alternative, more performant, SIMD based
+        /// implemetation.
+        OverrideJsonParseAndStringify(bool),
```
surma, line 247: Should this rather be called `SimdJsonBuiltins` or something along those lines?

crates/runner/src/lib.rs
```diff
@@ -52,6 +62,11 @@ impl Default for Builder {
             preload: None,
             command: JavyCommand::Build,
             redirect_stdout_to_stderr: None,
+            javy_stream_io: None,
```
surma, line 65: Not really related to this PR, but for future simplicity, but might be worth it to have the `Builder` autogenerated using [`derive-builder`](https://docs.rs/derive_builder/0.20.1/derive_builder/)

crates/cli/src/commands.rs
saulecabrera: Great suggestion, done!

Issue #768, opened 2024-10-01
RFC: Plugin support and dynamic linking changes for Javy


jeffcharles:
This is a proposal to add plugin support for Javy.

## What problems does it solve?

At the moment, if someone wants to add custom JS APIs to Javy, they need to fork most of the project (`javy-cli` and `javy-core`). Ideally it would be possible to add custom JS APIs to Javy without forking anything.

We also have a problem with dynamically linked modules where import namespaces generated by the Javy CLI are rolled out before support for those import namespaces is available where they would be used. Specifically, to support using dynamically linked modules in a given runtime environment, the runtime environment needs to configure a linker with one or more Javy import namespaces (e.g., `javy_quickjs_provider_v3`) to load a QuickJS provider Wasm module. New releases of the Javy CLI can potentially change the import namespace used (in the future, a release could introduce a `javy_quickjs_provider_v4` import namespace). However, a runtime environment using Javy would likely not be updated before the new release of the Javy CLI is available. So users using the latest Javy CLI release may create modules that are not compatible with that runtime environment.

As well, plugins should be able to specify their own import namespace because they may have their own JS APIs and the lifecycle of those APIs can be different than Javy CLI releases.

## What are the changes?

1. The `javy build` subcommand will accept a new `-C plugin=plugin.wasm` argument. The details of what's in the `plugin.wasm` file will be provided below.
2. The `-C plugin=plugin.wasm` argument will be _optional_ for the default statically linked modules. E.g., `javy build -C plugin=my_plugin.wasm`.
3. The `-C plugin=plugin.wasm` argument will be _required_ for the dynamically linked modules (this is a **breaking** change). E.g., `javy build -C dynamic -C plugin=my_plugin.wasm`.
4. An `initialize-plugin` subcommand will be added to the Javy CLI. This command will need to be run to initialize a plugin before it can be used. This is intended to be run prior to distributing the plugin and not on an end user's machine. E.g., `javy initialize-plugin path_to_myplugin.wasm -o=my_initialized_plugin.wasm`.
5. Change `emit-provider` to `emit-plugin` ~and apply `-J` options to the emitted plugin. E.g., `javy emit-plugin`~.

### Possible future changes

1. Update `emit-plugin` to accept `-J` options if there is demand for allowing runtime configurations on the vanilla plugin.

### Why these changes?

[...]

PR #776, opened 2024-10-02
Have provider `invoke` fn support not being given a function name

jeffcharles:
## Description of the change

Adds support to QuickJS provider's exported `invoke` function to handle not being given a function name. This shouldn't change any behaviour.

As part of #768, we'll want to remove the `eval_bytecode` exported function (to make it easier for plugin developers) and just have `invoke` so `invoke` needs to also be able to handle `eval_bytecode`'s use case where there is not an exported function to run.

I haven't deleted `eval_bytecode` in this PR because we still emit function invocations for it. I also haven't updated the dynamically linked codegen code to call `invoke` instead of `eval_bytecode` because all runtime environments using the provider would need to be updated with a provider version that includes this code change, otherwise new modules generated to use `invoke` with `0` arguments for function name pointer and length would fail.

## Checklist

### DIFF
--- a/crates/runner/src/lib.rs
+++ b/crates/runner/src/lib.rs
@@ -253,8 +253,7 @@ impl StoreContext {
 
 pub enum UseExportedFn {
     EvalBytecode,
-    Invoke,
-    InvokeWithFn(&'static str),
+    Invoke(Option<&'static str>),
 }
 
 impl Runner {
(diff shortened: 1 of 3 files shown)


crates/runner/src/lib.rs
```diff
@@ -251,6 +251,12 @@ impl StoreContext {
     }
 }
 
+pub enum UseExportedFn {
+    EvalBytecode,
+    Invoke,
+    InvokeWithFn(&'static str),
```
saulecabrera, line 257: Just a minor comment here - could this be consolidated into:

```rust
pub enum UseExportedFn {
  Invoke(Option<&'static str>)
  ...
}
```

saulecabrera (approved):
Left one minor comment -- not blocking, else LGTM.

crates/runner/src/lib.rs
jeffcharles: Yes, I can make that change.

PR #784, opened 2024-10-17
Inline function execution with bytecode evaluation

jeffcharles:
## Description of the change

Updates `run_bytecode` in javy-core to take an optional function name to execute and execute it if it's provided.

This substantially reduces the amount of fuel to call an exported function with dynamically linked modules. This also allows re-using a runtime after using it when calling `compile_src` (see #781).

In terms of performance numbers, given a function that looks like:
```js
function run() {
    console.log("hello");
}
run()
```

Fuel usage will increase from 36,929 to 36,983 (54 instructions).

Given a function that looks like:
```js
export function run() {
    console.log("hello");
}
```

Fuel usage will decrease from 92,080 to 40,283 (51,797 instructions).

I'm opting to not to update the `main.rs` file for the time being because I'm planning to have it deleted soon.

## Checklist

### DIFF
--- a/crates/core/src/execution.rs
+++ b/crates/core/src/execution.rs
@@ -3,7 +3,7 @@ use std::process;
 use anyhow::{anyhow, bail, Error, Result};
 use javy::{
     from_js_error,
-    quickjs::{context::EvalOptions, Ctx, Error as JSError, Function, Module, Value},
+    quickjs::{Ctx, Error as JSError, Function, Module, Value},
     to_js_error, Runtime,
 };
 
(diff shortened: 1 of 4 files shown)


crates/core/src/execution.rs
```diff
@@ -37,6 +46,7 @@ pub fn run_bytecode(runtime: &Runtime, bytecode: &[u8]) {
 /// the target function from a previously evaluated module. It's the caller's
 /// reponsibility to ensure that the module containing the target function has
 /// been previously evaluated.
+#[allow(dead_code)] // Used by `main.rs` but not by `lib.rs`.
```
saulecabrera, line 49: Any reason why we can't get rid of this function completely? I experimented a bit on top of your changes and it seems to work and have the same performance benefits as in the dynamically linked case:

```diff
diff --git a/crates/core/src/main.rs b/crates/core/src/main.rs
index 5f8db9f..1e8eb14 100644
--- a/crates/core/src/main.rs
+++ b/crates/core/src/main.rs
@@ -66,7 +66,7 @@ pub unsafe extern "C" fn invoke(fn_name_ptr: *mut u8, fn_name_size: usize) {
 
     let js_fn_name = str::from_utf8_unchecked(slice::from_raw_parts(fn_name_ptr, fn_name_size));
     let runtime = unsafe { RUNTIME.take().unwrap() };
-    execution::invoke_function(&runtime, FUNCTION_MODULE_NAME, js_fn_name);
+    execution::run_bytecode(&runtime, &BYTECODE.take().unwrap(), Some(js_fn_name));
 }
 
 // RAII abstraction for calling Wasm ctors and dtors for exported non-main functions.
```
I see some failures, all of them related to fuel consumption being lower than expected:

```
thread 'compile_test_exported_functions' panicked at crates/cli/tests/integration_test.rs:323:5:
fuel_consumed (63156) was not within 10.00% of the target_fuel value (80023)

```

saulecabrera (approved):
Very nice! Left a comment/suggestion. I don't think it's blocking, however, I think it's worth considering.

crates/core/src/execution.rs
jeffcharles: I was a little concerned that there might be a change in behaviour or a regression in fuel since the existing implementation wasn't evaluating the bytecode. But trying it out, it works and is faster, so let's go with it!

saulecabrera (approved):

PR #785, opened 2024-10-17
Disable reference-types since Wizer doesn't support them

jeffcharles:
## Description of the change

Disabling reference types when compiling javy-core.

The latest Rust stable compiler has started to emit instructions which rely on reference types support. Wizer (as of Wizer 7.0.0) [does not support reference types](https://github.com/bytecodealliance/wizer/blob/afcdfc702514fd8be98ddf025557297d2f79ae4c/src/lib.rs#L85). This causes the build for Javy CLI to fail [when Wizer is executed on javy-core](https://github.com/bytecodealliance/javy/actions/runs/11392246802/job/31698108529#step:6:317).

## Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,8 +2,15 @@
 # Cargo doesn't read directives in individual crates when invoking build
 # commands from the workspace root, hence adding it at the workspace root.
 # https://doc.rust-lang.org/cargo/reference/config.html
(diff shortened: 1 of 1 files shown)


saulecabrera (approved):
I was curious to know more about the concrete details of the change, for posterity and completeness here's a deeper explanation of of the inner workings of this change: https://github.com/WebAssembly/tool-conventions/issues/233

PR #786, opened 2024-10-18
Derive the import namespace from the provider module

jeffcharles:
## Description of the change

Updates the provider to have a custom section named `import_namespace` and has the Javy CLI use that custom section to determine what import namespace to use when creating dynamically linked modules.

Part of #768. This change will allow plugins to specify their own import namespaces instead of having to use one set in the Javy CLI. Also arguably, javy-core should be responsible for saying when to bump any version and not the CLI.

I had a little trouble figuring out a logical place to put the `derive_import_namespace_from_provider` function. I'm putting it in the `bytecode` module for now. Arguably we could introduce an abstraction around providers with various methods for things like getting a byte slice for writing to disk, performing bytecode compilation, and getting the import namespace. I'll probably need to do that refactoring very shortly to accommodate some changes I want to make to codegen so I'm thinking it makes sense to punt until I'm working on that.

## Checklist

### DIFF
--- a/crates/cli/src/codegen/builder.rs
+++ b/crates/cli/src/codegen/builder.rs
@@ -1,4 +1,7 @@
-use crate::codegen::{CodeGen, CodeGenType, DynamicGenerator, StaticGenerator};
+use crate::{
+    bytecode,
+    codegen::{CodeGen, CodeGenType, DynamicGenerator, StaticGenerator},
+};
 use anyhow::{bail, Result};
 use javy_config::Config;
 use std::path::PathBuf;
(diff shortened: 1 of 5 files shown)


crates/cli/src/codegen/builder.rs
```diff
@@ -44,11 +47,25 @@ impl WitOptions {
...
+    }
+}
+
 /// A code generation builder.
 #[derive(Default)]
 pub(crate) struct CodeGenBuilder {
-    /// The QuickJS provider module version.
-    provider_version: Option<&'static str>,
+    /// The import namespace for dynamically linked modules.
+    import_namespace: Option<ImportNamespace>,
```
saulecabrera, line 68: Is there a particular reason why this can't be `import_namespace: ImportNamespace` and default to `FromProvider`? i.e., drop the `Option`?

jeffcharles: Funnily enough that was my first implementation. Then I added back the option to bring it closer to how it's currently implemented on `main`. I'll drop the option and set the default to `FromProvider`.

jeffcharles: Actually, what I'll end up doing is having the provider do something polymorphically instead based on just merging #787.

PR #787, opened 2024-10-18
Refactor to use a Provider abstraction

jeffcharles:
## Description of the change

I'm adding a `Provider` abstraction which provides methods that need to be polymorphic over a provider.

As part of #768 I'll need to thread a plugin module through a bunch of the code. This refactoring front-loads some of those changes.

## Checklist


saulecabrera (approved):
Looks reasonable to me!

PR #794, opened 2024-10-29
Change JS config to use a schema and JSON

jeffcharles:
## Description of the change

This updates a few things relating to moving away from using a shared bitflag representation for JS configuration options between the Javy CLI, `javy-core`, and the Javy library crate:
- A new `shared_config` Cargo feature is added to the Javy library crate. Enabling this feature exports a Wasm function called `config_schema` that outputs a JSON representation of all of the configuration properties the provider supports. It also exports some Rust data structures and associated methods for use by the `javy-core` crate. These data structures and methods are used to parse a JSON data structure representing a JS configuration sent by the Javy CLI and apply the parsed representation to a pre-existing `Javy::Config` config.
- `javy-core` has been updated to enable the new `shared_config` feature and use its data structures to read in the JS configuration from the Javy CLI.
- The Javy CLI has been updated to obtain a list of supported JS configuration properties from using the `config_schema` Wasm function exported from the provider and to validate the `-J` flags against this schema. It has also been updated to encode the JS configuration as a JSON byte array when passing the JS configuration to the `initialize_runtime` function in the provider.
- The `javy-config` crate has been deleted which involved updating the Makefile and GitHub Action workflows.
- I moved some dependencies between the workspace and specific crates.

This change would not support runtime configurations that are not defined in the Javy crate. This change also only supports boolean configurations. It should be possible to augment the approach we're using if there is a need to in the future.

Part of #768. Since plugins will exist outside of the Javy source tree, they cannot depend on the unpublished `javy-config` crate. Also it's easier for bitflags to get out of sync betweeen different source trees and when the bitflags get out of sync, the error messages that result are not helpful. So, we need a way for the Javy CLI to know what configuration properties a plugin supports along with associated help text. We also need a way for the CLI and plugin to understand each other when a configuration is passed into `initialize_runtime`. JSON with using UTF-8 strings to describe the configuration properties seems like a relatively straightforward approach.

## Checklist


jeffcharles:
I'm also updating the `Makefile` for `test-javy` to match what we do in CI.

saulecabrera (commented):
Left one comment, I'm mostly trying to reconcile these changes with our conversation earlier today. This is a sketch of my understanding from our discussion:

* Vanilla plugin: lives in tree, and it's a generic JS environment. This is not our current `core` crate.
* When specifying `-C dynamic` and/or `-C plugin` not `-J` options are specified (this is the most simple route). We expect, at least initially or until someone actually needs this functionality that the plugin's `initialize_runtime` will take care of initializing the JS runtime according to the plugin's needs. So no `-J` options are needed immediately. 
* `-J` _could_ be allowed for the vanilla plugin, which we don't have today and don't have to introduce immediately. 
* Ultimately our goal, at least in my head, is to remove the `core` crate entirely from Javy itself and treat it as plugin moving forward. 


It's totally possible that I misunderstood some of our discussion points, so thanks for bearing with me. 

In general, if my sketch above is not totally off, I think we can keep most of the changes here, except for the shared config parsing, which would involve eliminating parsing the `-J` group from the CLI.

jeffcharles:
Discussed offline. The current plan is to make JS runtime configuration flags (that is, `-J` flags) and a future plugin flag mutually exclusive. As a result, the `SharedConfig` introduced in this PR and the associated Wasm export and parsing logic can be removed from the Javy library crate and placed in the `javy-core` crate.

saulecabrera (approved):

PR #795, opened 2024-10-30
Create plugin-api crate and move most of core's logic into it

jeffcharles:
## Description of the change

Adds a library crate named `javy-plugin-api` with a lot `javy-core`'s logic. The idea is a Javy plugin developer should make their Javy plugin cdylib dependent on this `javy-plugin-api` crate and it will handle most of the exports for them and let them configure their runtime.

Part of #768. We don't want every plugin developer to have to re-implement most of the existing logic in `javy-core` themselves so this provides a crate they can depend on so they don't have to re-implement the existing logic.

I'm opting to leave the allocation function definitions in the `javy` library crate for this PR. I'll move the functions in the next PR.

## Checklist


crates/core/Cargo.toml
```diff
@@ -11,10 +11,9 @@ crate-type = ["cdylib"]
 
 [dependencies]
 anyhow = { workspace = true }
-javy = { workspace = true, features = ["export_alloc_fns", "json"] }
-once_cell = { workspace = true }
+javy-plugin-api = { workspace = true, features = ["json"] }
```
saulecabrera, line 14: Is it intended to have this dependency at the workspace level? I think it's not shared across any other crates in the workspace, right?

crates/plugin-api/src/lib.rs
```diff
@@ -0,0 +1,220 @@
+//! A crate for creating Javy plugins
+//!
+//! Example usage:
+//! ```
```
saulecabrera, line 4:
```suggestion
//! ```rust
```

saulecabrera: To add syntax highlighting in LSP-powered highlighting.

saulecabrera (approved):
A couple of nits, but overall LGTM!

crates/core/Cargo.toml
jeffcharles: I can move it out of the workspace into `javy-core`. For some reason I thought the `javy` crate was only used by `javy-core` but I can see after looking that the fuzzer also uses it.

PR #797, opened 2024-10-31
Rename javy-core to javy-plugin

jeffcharles:
## Description of the change

Renaming `javy-core` to `javy-plugin`.

I've updated the name in our documentation but I will go through the documentation more holistically in the future to ensure it's up-to-date as opposed to doing that as part of this PR. I have also tested the release GitHub Action to determine that it is working correctly with this change.

Part of #768. The `javy-core` crate is the default plugin for Javy so a name like `javy-plugin` better reflects that.

## Checklist


saulecabrera (approved):
Looks good to me -- the `test_core` step in CI seems stuck, I think it's due to the name of the jobs?

jeffcharles:
It's because the `test_core` check is marked as mandatory but the name of the workflow has been changed to `test_plugin`. So I need to remove `test_core` as a mandatory check before I can merge and then I can mark `test_plugin` as mandatory.

jeffcharles:
Branch protection rules have been updated

PR #813, opened 2024-11-04
Allow plugins when building static modules

jeffcharles:
## Description of the change

Add a `-p` or `--plugin` flag to `javy build` to support using a user-provided plugin. I've disabled allowing plugins on dynamically linked modules in this PR since I want to make that a required parameter and that would represent a breaking change. I also don't want a version of Javy shipped where it is an optional parameter.

I've also added a test-plugin crate for use by the CLI tests so the tests can verify the plugin is being actually used by the generated Wasm module.

Part of #768. This enables using plugins for statically linked modules.

I opted to compile the test plugin as part of the Makefile and CI steps but I'm open to switching it so we commit the final artifact if that would be preferrable.

## Checklist

### DIFF
--- a/crates/cli/src/codegen/mod.rs
+++ b/crates/cli/src/codegen/mod.rs
@@ -279,7 +279,7 @@ impl Generator {
             // tests are executed with Javy built with the release profile so
             // `debug_assert!`s are stripped out.
             assert!(
-                matches!(self.plugin, Plugin::User { .. }),
+                self.plugin.is_user_plugin(),
                 "Using invoke with null function only supported for user plugins"
             );
             instructions
@@ -358,9 +358,9 @@ impl Generator {
             CodeGenType::Static => {
                 // Remove no longer necessary exports.
                 module.exports.remove("canonical_abi_realloc")?;
-                // `eval_bytecode` is not present in user plugins and `remove`
-                // returns an error if the export is not present.
-                if module.exports.get_func("eval_bytecode").is_ok() {
+                // User plugins won't have an `eval_bytecode` function that
+                // Javy "owns".
+                if !self.plugin.is_user_plugin() {
                     module.exports.remove("eval_bytecode")?;
                 }
                 module.exports.remove("invoke")?;
--- a/crates/cli/src/codegen/mod.rs
+++ b/crates/cli/src/codegen/mod.rs
@@ -66,7 +66,7 @@ pub(crate) enum CodeGenType {
 // This is an internal detail of this module.
 pub(crate) struct Identifiers {
     canonical_abi_realloc: FunctionId,
-    eval_bytecode: FunctionId,
+    eval_bytecode: Option<FunctionId>,
     invoke: FunctionId,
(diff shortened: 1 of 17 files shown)


crates/cli/src/codegen/mod.rs
```diff
@@ -259,11 +260,35 @@ impl Generator {
...
+            // invoke for `javy_quickjs_provider_v*`.
+            // `javy_quickjs_provider_v2` will never support calling `invoke`
+            // with a null function. Older `javy_quickjs_provider_v3`'s do not
+            // support being called with a null function. User plugins and
+            // newer `javy_quickjs_provider_v3`s do support being called with a
+            // null function.
+            // Using `assert!` instead of `debug_assert!` because integration
+            // tests are executed with Javy built with the release profile so
+            // `debug_assert!`s are stripped out.
+            assert!(
```
jeffcharles, line 281: My plan is to relax this assertion a bit after we have released a new version of `javy_quickjs_provider_v3`.

crates/cli/src/codegen/mod.rs
```diff
@@ -333,7 +358,11 @@ impl Generator {
             CodeGenType::Static => {
                 // Remove no longer necessary exports.
                 module.exports.remove("canonical_abi_realloc")?;
-                module.exports.remove("eval_bytecode")?;
+                // `eval_bytecode` is not present in user plugins and `remove`
+                // returns an error if the export is not present.
+                if module.exports.get_func("eval_bytecode").is_ok() {
+                    module.exports.remove("eval_bytecode")?;
+                }
```
saulecabrera, line 365: This is true because we know the plugin API surface, however, I don't think anything is preventing a plugin from exporting a function named `eval_bytecode` for their own purposes. Probably unlikely, but not impossible. I'd suggest checking if the plugin is not a user provided plugin before removing this export.

crates/cli/src/commands.rs
```diff
@@ -105,6 +105,10 @@ pub struct BuildCommandOpts {
     /// JavaScript runtime options.
     /// Use `-J help` for more details.
     pub js: Vec<JsGroupValue>,
+
+    #[arg(short, long)]
+    /// Path of the plugin.
+    pub plugin: Option<PathBuf>,
```
saulecabrera, line 111: Any reason why the plugin flag is not scoped? [In the _What are the changes?_ section of the RFC](https://github.com/bytecodealliance/javy/issues/768), the plugin handling is scoped under the `-C` option group, which is what I think makes most sense, since it's easier to validate the semantic meaning of the group as a whole (e.g., `-C pluing=..` is required with `-C dynamic`)

saulecabrera (commented):
Looking good, left a few comments.

crates/cli/src/codegen/mod.rs
jeffcharles: I've updated the check

saulecabrera (approved):
Thanks!

PR #818, opened 2024-11-06
Simplify dynamic linking imports tests

jeffcharles:
## Description of the change

Simplifies the test code for checking the imports that dynamically linked modules have. At the moment, we always import `canonical_abi_realloc`, `memory`, `eval_bytecode`, and `invoke`, so we don't need separate tests for dynamically linked modules that have named exports.

I'm going to need to adjust how we check the imports to exclude `eval_bytecode` and reduce the number of expected imports for certain dynamically linked modules in an upcoming PR and wanted to refactor the test suite outside of that PR first.

## Checklist

### DIFF
--- a/crates/runner/src/lib.rs
+++ b/crates/runner/src/lib.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, bail, Result};
+use anyhow::{bail, Result};
 use std::error::Error;
 use std::fmt::{self, Display, Formatter};
 use std::io::{self, Cursor, Write};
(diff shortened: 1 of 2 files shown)


crates/runner/src/lib.rs
```diff
@@ -432,63 +432,49 @@ impl Runner {
         let module = Module::from_binary(self.linker.engine(), &self.wasm)?;
```
saulecabrera, line 432: As a nit, could you given the description of the PR, could you rename this function reflect the intention? e.g., `ensure_expected_imports`?

crates/runner/src/lib.rs
```diff
@@ -432,63 +432,49 @@ impl Runner {
...
+
+        let invoke = imports
+            .iter()
+            .find(|i| i.name() == "invoke")
+            .expect("Should have eval_bytecode import");
+        let ty = invoke.ty();
+        let f = ty.unwrap_func();
+        assert!(f.params().all(|p| p.is_i32()));
+        assert_eq!(4, f.params().len());
+        assert_eq!(0, f.results().len());
```
saulecabrera, line 475: Could you replace the `assert`s with `bail!` or `Err(anyhow::Error)`? I get that we're asserting and that the runner is mostly used in tests, however, I'd like to ensure that errors are recoverable here so that failures can bubble up with as much as context as possible to the test invoking them.

saulecabrera (approved):
Thanks!

PR #821, opened 2024-11-07
Add support for plugins to dynamically linked modules

jeffcharles:
## Description of the change

Changes from not supporting a plugin when creating dynamically linked modules to requiring a plugin when creating dynamically linked modules. This is a breaking change.

I've also updated the dynamic codegen to not emit calls to `eval_bytecode` for the default plugin.

Part of #768. We want to support using plugins when building dynamically linked modules. And, as I mention in a comment, we never want to assume the import namespace in the Javy CLI, so we need a plugin to be specified so we can derive the import namespace to use from it.

For the change to stop emitting calls to `eval_bytecode`, we want to remove `eval_bytecode` from the default plugin and the first step to doing that is to stop emitting new calls to `eval_bytecode`. The requirement to specify a plugin when building new dynamically linked modules means we can be sure the plugin used in any execution environment also supports calling `invoke` with a null function assuming the same or a newer plugin is used in the execution environment as is used when invoking `javy build -C dynamic -C plugin=...`.

## Checklist

### DIFF
--- a/docs/docs-using-nodejs.md
+++ b/docs/docs-using-nodejs.md
@@ -11,15 +11,15 @@ This example shows how to use a dynamically linked Javy compiled Wasm module. We
 
 ### Steps
 
-1. The first step is to compile the `embedded.js` with Javy using dynamic linking:
-```shell
-javy build -C dynamic -o embedded.wasm embedded.js
-```
-2. Next emit the Javy plugin
+1. Emit the Javy plugin
 ```shell
 javy emit-provider -o plugin.wasm
 ```
-3. Then we can run `host.mjs`
+2. Compile the `embedded.js` with Javy using dynamic linking:
+```shell
+javy build -C dynamic -C plugin=plugin.wasm -o embedded.wasm embedded.js
+```
+3. Run `host.mjs`
 ```shell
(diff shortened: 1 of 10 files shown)


docs/docs-using-nodejs.md
```diff
@@ -11,15 +11,15 @@ This example shows how to use a dynamically linked Javy compiled Wasm module. We
 
 ### Steps
 
-1. The first step is to compile the `embedded.js` with Javy using dynamic linking:
+1. Emit the Javy plugin
 ```shell
-javy build -C dynamic -o embedded.wasm embedded.js
+javy emit-provider -o plugin.wasm
```
saulecabrera, line 16: Should this be `emit-plugin`?

saulecabrera (approved):

docs/docs-using-nodejs.md
jeffcharles: I haven't merged the PR where I rename the command yet. I assume this will register as a merge conflict if I merge that PR before this one or if I merge this PR and then try to merge that PR because PRs are changing the same line in different ways.

jeffcharles: And it looks like it did register as a merge conflict 🎉

Issue #70, opened 2022-01-26
CMake is undocumented dependency for local build


nickwesselman:
I was unable to successfully run `make` without installing [CMake](https://cmake.org/):

```
error: failed to run custom build command for `binaryen-sys v0.12.0`

Caused by:
  process didn't exit successfully: `/Users/nickwesselman/src/javy/target/release/build/binaryen-sys-fd3d87b2d3e0fe08/build-script-build` (exit status: 101)
  --- stdout
  running: "cmake" "/Users/nickwesselman/.cargo/registry/src/github.com-1ecc6299db9ec823/binaryen-sys-0.12.0/binaryen" "-DBUILD_STATIC_LIB=ON" "-DENABLE_WERROR=OFF" "-DCMAKE_INSTALL_PREFIX=/Users/nickwesselman/src/javy/target/release/build/binaryen-sys-1ef0729e257a20ee/out" "-DCMAKE_C_FLAGS= -ffunction-sections -fdata-sections -fPIC -arch arm64" "-DCMAKE_C_COMPILER=/usr/bin/cc" "-DCMAKE_CXX_FLAGS= -ffunction-sections -fdata-sections -fPIC -arch arm64" "-DCMAKE_CXX_COMPILER=/usr/bin/c++" "-DCMAKE_ASM_FLAGS= -ffunction-sections -fdata-sections -fPIC -arch arm64" "-DCMAKE_ASM_COMPILER=/usr/bin/cc" "-DCMAKE_BUILD_TYPE=Release"

  --- stderr
  fatal: not a git repository (or any of the parent directories): .git
  thread 'main' panicked at '
  failed to execute command: No such file or directory (os error 2)
  is `cmake` not installed?

  build script failed, must exit now', /Users/nickwesselman/.cargo/registry/src/github.com-1ecc6299db9ec823/cmake-0.1.45/src/lib.rs:894:5
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Should this be documented in the README? Happy to PR if so.

saulecabrera:
Good catch, it seems that `cmake` is no longer installed by default in M1 macs; this was not an issue in Intel ones. But yeah, we should add this to the development section of the README

PR #829, opened 2024-11-18
Remove cmake from build requirements

jeffcharles:
## Description of the change

Removing `cmake` from the list of things that need to be installed to build Javy.

`cmake` does not appear to be necessary. Or at least I was able to build Javy on a fresh Ubuntu 24.04 without it.

## Checklist

### DIFF
--- a/docs/docs-contributing-building.md
+++ b/docs/docs-contributing-building.md
@@ -4,9 +4,6 @@
 - [rustup](https://rustup.rs/)
 - Stable Rust, installed via `rustup install stable && rustup default stable`
 - wasm32-wasip1, can be installed via `rustup target add wasm32-wasip1`
-- cmake, depending on your operating system and architecture, it might not be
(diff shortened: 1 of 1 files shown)


saulecabrera (commented):
Do you remember what introduced the need for cmake? I'm just wondering if the root cause is  known and we can confidently remove this dependency from the docs. If we're no confident about the root cause, perhaps we can consider keeping this around? Going back in history, it seems that [binaryen-rs is what introduced the cmake dependency](https://github.com/bytecodealliance/javy/issues/70), we have now migrated to wasm-opt-rs, so I _think_ we should be good.

jeffcharles:
It looks like it was `binaryen-rs` to me as well. FWIW, a fresh docker image of Ubuntu 24.04 does not have `cmake` installed and Javy will build on it without any errors. I'm skeptical that there would be some dependency that would require `cmake` but _only_ on MacOS.

saulecabrera (approved):

PR #836, opened 2024-11-19
Adjust fuzzing configs to match what's in QuickJS

jeffcharles:
## Description of the change

Adding some parameters to the Javy library crate's `Config` so we can set them to non-default values when fuzzing. Also updating the fuzzer to set those parameters. Also throwing any fuzzing inputs with more than 350 `{` characters.

Note that the fuzzer still crashes with out of memory errors with this change.

We see out of memory errors when running the fuzzer continuously. This updates a few configs to match what's in QuickJS's fuzzer. I also don't try to run strings with more than 350 `{` characters because that trips the maximum stack size. We should consider investigating that further but I'd like to at least get the fuzzer running continuously.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1533,7 +1533,7 @@ dependencies = [
 
 [[package]]
 name = "javy"
-version = "3.0.3-alpha.1"
+version = "3.1.0-alpha.1"
 dependencies = [
  "anyhow",
  "bitflags",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ wasmtime = "23"
 wasmtime-wasi = "23"
 wasi-common = "23"
 anyhow = "1.0"
-javy = { path = "crates/javy", version = "3.0.3-alpha.1" }
+javy = { path = "crates/javy", version = "3.1.0-alpha.1" }
 tempfile = "3.13.0"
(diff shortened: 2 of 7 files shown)


saulecabrera (approved):
Regarding the OOM -- I wonder if the usage of `ManuallyDrop` inside the runtime is at play here? Have you investigated removing its usage? What I'm getting at is that we never recycle objects that are no longer referenced inside the runtime since the [context is behind a `ManuallyDrop`.](https://github.com/bytecodealliance/javy/blob/a79795216de32fea21aee4ef84bf1733372b5cf5/crates/javy/src/runtime.rs#L36)

jeffcharles:
We never drop the runtime while fuzzing since they're inside mutable statics so the drops on the rquickjs runtime and context will never run during fuzzing anyway.

I did try switching to building a new runtime everytime and removing the manual drops and that seemed to slow down the memory growth but I was still seeing out of memory errors.

jeffcharles:
Actually trying that again with this branch, I get a hard crash on `Assertion failed: (list_empty(&rt->gc_obj_list)), function JS_FreeRuntime, file quickjs.c, line 2002.`. I'm guessing that's from the JSON functions still being in the `gc_obj_list` since that was what I saw with `dump-leaks` on the other PR.

PR #838, opened 2024-11-22
Make event loop a runtime config instead of a Cargo feature

jeffcharles:
## Description of the change

Changes `experimental_event_loop` from a Cargo feature to a runtime config. This also introduces a `javy_plugin_api::Config` that wraps the `javy::Config` so we can add configuration options that only affect the `javy_plugin_api` crate behaviour and that don't inherently change how the `javy::Runtime` behave (that's why I didn't add it to `javy::Config`). Since we change the type that needs to be passed for `initialize_runtime`, I've bumped the `javy-plugin-api` version by a major version. We have minimal users of the plugin API crate so the impact should be limited.

This should simplify the build and test infrastructure since we only need to build one artifact for the plugin and the CLI tests instead of two plugins. This should also help with cacheability of build artifacts. This will also make it easier to use the event loop functionality.

## Checklist


jeffcharles:
The _Test CLI Features_ check will have to be marked as not required to merge this since this PR deletes that workflow.

saulecabrera (approved):
Changes look good to me. Would it make sense to consider renaming the the option to `allow-event-loop` or something along those lines? It'd be ideal to signal that it's no longer considered experimental, for the same reasons exposed in: https://github.com/bytecodealliance/javy/pull/833

saulecabrera (approved):

PR #852, opened 2024-12-04
Tweak Dependabot to not group QuickJS + ignore wasmtime

jeffcharles:
## Description of the change

Tweaks Dependabot configuration to do two things:
- Stop grouping rquickjs updates in with the non-breaking updates. We have to be very careful with these so they should be inspected separately.
- Ignore Wasmtime updates. Wizer is generally going to pin Wasmtime to a particular version.

Trying to get on top of the Dependabots opened this month and one of them is for rquickjs which I want to evaluate separately.

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,9 +7,13 @@ updates:
     open-pull-requests-limit: 100
     groups:
       nonbreaking:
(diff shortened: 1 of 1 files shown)


saulecabrera (approved):
Good call.

PR #879, opened 2025-01-21
Plugin API Surface Refactor

ianmarmour:
## Description of the change

These changes clean up the interfaces around code generation and plugins so that they will be ready to spin off into their own crate without dramatically increasing the API surface of the new crate. The overview of the changes made are below,

- Updated - `Plugin` is now a wrapper around a `Vec<u8>`.
- Updated - `Generator` now stores `plugin_kind` internally to determine if it's linking to a User/Default/V2 `Plugin`.
- Updated - `Generator` now exposes `linking_v2_plugin` or `linking_default_plugin` that can have their visibility controlled with a future feature flag to support the existing `cli` use cases.
- Updated - Some files were moved around to ensure they're in the `codegen` crate rather then `cli`.
- Added - `CliPlugin` wrapper that stores `PluginKind` information for use with config generation logic in the `cli`.
- Removed - `CodeGenBuilder` was removed as it's behavior is provided by `Generator`.

To lower the API surface of the plugin/code-generation components in preparation for [a future split of that logic into a `javy-codegen`](https://github.com/bytecodealliance/javy/pull/877) crate based on @jeffcharles request.

## Checklist

- [ ] I've updated the relevant CHANGELOG files if necessary. Changes to `javy-cli` and `javy-plugin` do not require updating CHANGELOG files.


crates/cli/src/codegen/mod.rs
```diff
@@ -370,11 +401,12 @@ impl Generator {
             CodeGenType::Static => {
                 // Remove no longer necessary exports.
                 module.exports.remove("canonical_abi_realloc")?;
-                // User plugins won't have an `eval_bytecode` function that
-                // Javy "owns".
-                if !self.plugin.is_user_plugin() {
+
+                // Only internal plugins expose eval_bytecode function.
+                if matches!(self.plugin.kind(), PluginKind::Internal(..)) {
```
jeffcharles, line 406: Similarly here, I was hoping this would look more like:

```rust
if self.linking_v2_plugin || self.linking_default_plugin {
```

with a similar setter for `linking_default_plugin` that we could gate behind a feature.

crates/cli/src/plugins.rs
```diff
@@ -1,86 +1,113 @@
...
 
-impl Default for Plugin {
-    fn default() -> Self {
-        Self::Default
-    }
+/// Represents any valid Javy plugin.
+#[derive(Clone, Debug)]
+pub(crate) struct Plugin {
+    pub(crate) bytes: Vec<u8>,
+    kind: PluginKind,
```
jeffcharles, line 56: Instead of this, can we do:

```suggestion
/// Javy plugin.
pub struct Plugin(Vec<u8>);

// Represents any valid Javy plugin.
#[derive(Clone, Debug)]
pub(crate) struct CliPlugin {
    pub(crate) bytes: Plugin,
    kind: PluginKind,
```

With the idea that the `Plugin` struct will live in the codegen crate and the `CliPlugin` will live in the CLI crate the `Generator` will reference `Plugin`. Basically I want to keep the `PluginKind` knowledge entirely in the CLI crate and use a couple feature gated booleans to get `Generator` to do what it needs to do rather than use a richer plugin abstraction in the generator so it's straightforward for a codegen user to just pass their plugin bytes along without having to worry about `PluginKind`s.

crates/cli/Cargo.toml
```diff
@@ -11,7 +11,9 @@ name = "javy"
 path = "src/main.rs"
 
 [features]
+default = ["plugin-internal"]
 dump_wat = ["dep:wasmprinter"]
+plugin-internal = []
```
jeffcharles, line 16: I don't want to introduce the feature gates in this PR but rather when the codegen crate is created. That said, it would be helpful to mark off with PR comments as to which methods would be feature gated.

crates/cli/src/js_config.rs
```diff
@@ -1,5 +1,72 @@
...
+use wasmtime::{AsContextMut, Engine, Linker};
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct ConfigSchema {
+    pub(crate) supported_properties: Vec<JsConfigProperty>,
+}
+
+impl ConfigSchema {
+    pub(crate) fn from_plugin(plugin: &Plugin) -> Result<Option<ConfigSchema>> {
```
jeffcharles, line 19: I think you can keep this method on `CliPlugin`, just not the codegen plugin.

jeffcharles (commented):
Thanks for the work! I understand it takes time to do and I appreciate you taking the time to do it.

That said, I do have concerns about what's in here. I think there was some misunderstanding around the abstractions I would've liked to see for the `Plugin`. Specifically, I wanted a richer `Plugin` struct/abstraction for use in the CLI crate and a _different_ and very slimmed down `Plugin` struct/abstraction for use in the future codegen crate (specifically in the `Generator`) with additional properties and gated setters on the `Generator` to preserve behaviour with a very slimmed down plugin abstraction.

Does that make sense?

crates/cli/src/codegen/mod.rs
```diff
@@ -119,28 +121,51 @@ pub(crate) struct Generator {
...
     /// WIT options for code generation.
     pub wit_opts: WitOptions,
+    /// An optional JS runtime config provided as JSON bytes.
+    js_runtime_config: Option<Vec<u8>>,
+}
+
+#[cfg(feature = "plugin-internal")]
+impl Generator {
+    /// Creates a new [`Generator`].
+    pub(crate) fn new(ty: CodeGenType, js_runtime_config: Vec<u8>, plugin: Plugin) -> Self {
```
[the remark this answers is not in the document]
ianmarmour: What's the purpose of having the `CodeGenBuilder` and then having the `Generator` itself also be a builder? That seems fairly redundant to me...?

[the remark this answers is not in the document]
jeffcharles: You're right, I overlooked that. I'm fine with having a single `new` function for `Generator` (that is, it has the same parameters whether the feature is enabled or disabled) with the booleans indicating which linking logic to use given the new function shouldn't be public outside the crate and having the builder have gated public functions for setting those booleans.

crates/cli/src/codegen/builder.rs
ianmarmour, line 1: I couldn't see a use case where this was required over just implementing the generator itself as a builder pattern.

crates/cli/src/codegen/bytecode.rs
```diff
@@ -2,20 +2,18 @@ use anyhow::{anyhow, Result};
 use wasi_common::{sync::WasiCtxBuilder, WasiCtx};
 use wasmtime::{AsContextMut, Engine, Instance, Linker, Memory, Module, Store};
 
-use crate::plugins::Plugin;
-
-pub fn compile_source(plugin: &Plugin, js_source_code: &[u8]) -> Result<Vec<u8>> {
-    let (mut store, instance, memory) = create_wasm_env(plugin)?;
+pub(crate) fn compile_source(plugin_bytes: &[u8], js_source_code: &[u8]) -> Result<Vec<u8>> {
```
ianmarmour, line 5: Happy to change these back if we want to.

crates/cli/src/codegen/mod.rs
```diff
@@ -106,43 +111,88 @@ impl BytecodeMetadata {
...
     /// Whether to embed the compressed JS source in the generated module.
-    pub source_compression: bool,
+    pub(crate) source_compression: bool,
     /// WIT options for code generation.
-    pub wit_opts: WitOptions,
+    pub(crate) wit_opts: wit::WitOptions,
+    /// JavaScript function exports.
+    pub(crate) function_exports: exports::Exports,
+    /// The kind of plugin a generator will link.
+    plugin_kind: plugin::PluginKind,
```
ianmarmour, line 127: This will never be exposed directly to customers. I know you mentioned removing the `PluginKind` entirely from the `codegen` crate but storing boolean flags for what we're linking against is actually more complex. I think this solution of abstracting the `PluginKind` behind boolean setters helps create the interface you're looking for while slightly reducing the implementation complexity (mostly around error handling and checking if each boolean is set).

crates/cli/src/codegen/mod.rs
```diff
@@ -106,43 +111,88 @@ impl BytecodeMetadata {
...
+
+        self
+    }
+
+    /// Runtime configuration options to pass to the module.
+    pub fn js_runtime_config(&mut self, js_runtime_config: Vec<u8>) -> &mut Self {
+        self.js_runtime_config = js_runtime_config;
+        self
+    }
+}
```
ianmarmour, line 185: These will all sit behind the feature flag.

crates/cli/src/codegen/mod.rs
```diff
@@ -197,8 +247,15 @@ impl Generator {
...
-            CodeGenType::Dynamic => {
-                let import_namespace = self.plugin.import_namespace()?;
+            LinkingKind::Dynamic => {
+                // All plugins by default are assumed to be linking against default
+                // or no module. For plugins linking against the V2 module we need
+                // the import_namespace will always be hardcoded to the same value.
+                let import_namespace = match self.plugin_kind {
+                    plugin::PluginKind::V2 => "javy_quickjs_provider_v2".to_string(),
+                    _ => self.plugin.import_namespace()?,
+                };
```
ianmarmour, line 257: So this kind of sucks... I'll admit that but I couldn't think of a better way to keep the `import_namespace` function in this crate while supporting the existing behavior/tests.

crates/cli/src/codegen/plugin.rs
```diff
@@ -1,147 +1,81 @@
...
+#[derive(Default, Clone)]
+pub(crate) enum PluginKind {
+    #[default]
+    None,
     Default,
-    /// A plugin for use with the `compile` to maintain backward compatibility.
     V2,
-    /// A plugin provided by the user.
-    User { bytes: Vec<u8> },
+}
```
ianmarmour, line 20: This could probably just be moved to mod as it's only consumers are in there... but it did feel a bit awkward not to have it live alongside `Plugin`

crates/cli/src/plugin.rs
```diff
@@ -0,0 +1,31 @@
+use crate::codegen::plugin::Plugin;
```
ianmarmour, line 1: I don't love this... maybe I would hate it less after we split out the `codegen` crate but at the moment it feels really awkward to have two basically identical structs laying around. Open to suggests here we really only need this wrapper for generating the `JsConfig` :|.

crates/cli/src/codegen/plugin.rs
```diff
@@ -1,147 +1,81 @@
...
-            Self::User { bytes } => bytes,
-        }
+        self.bytes.as_slice()
     }
+}
 
-    /// Uses the plugin to generate QuickJS bytecode.
+impl Plugin {
+    /// Uses a plugin to generate QuickJS bytecode.
     pub fn compile_source(&self, js_source_code: &[u8]) -> Result<Vec<u8>> {
```
ianmarmour, line 54: This probably doesn't matter quite yet but do we want/need this to be public? Do we not just want the `Generators` generate method to be public instead?

crates/cli/src/codegen/mod.rs
jeffcharles: I'm totally okay with having `PluginKind` in the generator if it's not exposed in the public API used by most consumers 👍

crates/cli/src/codegen/mod.rs
jeffcharles: Could you add a function like `plugin_kind.import_namespace(&plugin) -> String` to `PluginKind` with this implementation? It's moving logic around but at least that way the namespace will be in the `plugin` module as opposed to the generator.

crates/cli/src/codegen/plugin.rs
jeffcharles: I'm actually good with it being in this module 👍

crates/cli/src/plugin.rs
jeffcharles: Yeah it's a little awkward at the moment but it'll help shrink the size of the next change so I'm in favour.

crates/cli/src/codegen/plugin.rs
jeffcharles: Yeah a bunch of the methods that were public probably should be changed to `pub(super)` or `pub(crate)`.

crates/cli/src/codegen/mod.rs
```diff
@@ -50,18 +50,23 @@ use wasm_opt::{OptimizationOptions, ShrinkLevel};
 use wasmtime_wasi::{pipe::MemoryInputPipe, WasiCtxBuilder};
 use wizer::{Linker, Wizer};
 
-use crate::{js_config::JsConfig, plugins::Plugin, JS};
 use anyhow::Result;
 
 static STDIN_PIPE: OnceLock<MemoryInputPipe> = OnceLock::new();
 
-pub(crate) enum CodeGenType {
-    /// Static code generation.
```
jeffcharles, line 59: It's admittedly a little silly but I'd like we have triple slash comments on anything that'll be public. There are clippy settings and other lint things that flag missing triple-slash comments on public parts of the API. And this will at some point be public again. Granted the comment will need to be updated to reflect the new enum name.

crates/cli/src/codegen/plugin.rs
```diff
@@ -1,147 +1,81 @@
...
-    /// A plugin for use with the `compile` to maintain backward compatibility.
     V2,
-    /// A plugin provided by the user.
-    User { bytes: Vec<u8> },
+}
+
+/// Represents any valid Javy plugin.
+#[derive(Clone, Debug)]
+pub struct Plugin {
+    pub bytes: Vec<u8>,
```
jeffcharles, line 25: Can we remove the `pub`? The public `.as_bytes()` method should offer accessibility to get at underlying bytes.

crates/cli/src/codegen/plugin.rs
```diff
@@ -1,147 +1,81 @@
...
+            })
+            .ok_or_else(|| anyhow!("Plugin is missing import_namespace custom section"))?
+            .data(&Default::default()); // Argument is required but not actually used for anything.
+        Ok(str::from_utf8(&import_namespace)?.to_string())
     }
 }
 
 /// A validated but uninitialized plugin.
-pub(super) struct UninitializedPlugin<'a> {
+pub struct UninitializedPlugin<'a> {
```
jeffcharles, line 78: Do you need the plugin initialization logic in a crate? I'd prefer to keep that logic in the CLI, it's not a part of the codegen process, it's just used by `javy init-plugin`.

crates/cli/src/main.rs
```diff
@@ -104,6 +120,6 @@ fn emit_plugin(opts: &EmitPluginCommandOpts) -> Result<()> {
         Some(path) => Box::new(File::create(path)?),
         _ => Box::new(std::io::stdout()),
     };
-    file.write_all(Plugin::Default.as_bytes())?;
+    file.write_all(Plugin::new(PLUGIN_MODULE.to_vec()).as_bytes())?;
```
jeffcharles, line 123: Nit:

```suggestion
    file.write_all(PLUGIN_MODULE)?;
```

crates/cli/src/codegen/plugin.rs
```diff
@@ -1,147 +1,81 @@
...
-
-/// Different plugins that are available.
-#[derive(Debug)]
-pub enum Plugin {
-    /// The default plugin.
+/// Represents what kind of plugin.
+#[derive(Default, Clone)]
+pub(crate) enum PluginKind {
+    #[default]
+    None,
```
jeffcharles, line 17:
```suggestion
    User,
```

Same with the CLI enum. `None` would suggest to me that there isn't a plugin being used but there is.

jeffcharles (commented):
Mostly looks good! Just would like to figure out where the plugin initialization logic should live and fix up some minor things but otherwise it's what I'm looking for structurally.

crates/cli/src/codegen/mod.rs
ianmarmour: I like that quite a bit more.

crates/cli/src/codegen/plugin.rs
ianmarmour: I don't need it in the create, lets keep it in the CLI.

jeffcharles (approved):
Thank you for your hard work and patience! LGTM!

jeffcharles:
Actually I do need you to run `cargo fmt` and `cargo clippy` on all crates. The tests passed but linting failed. And then this should be good to go.

jeffcharles (approved):
LGTM

PR #1010, opened 2025-09-04
Allow static mut ref in fuzzing to get lint passing

jeffcharles:
## Description of the change

Stop warning on the lint failure in the fuzzing.

If we ever start linting this file, not doing this results in Clippy failing.

## Checklist


saulecabrera (commented):
> If we ever start linting this file

Any reason why we aren't? Could we include this file for linting as part of this change?

PR #1023, opened 2025-10-03
Add string_normalize intrinsic

jeffcharles:
## Description of the change

Adds the `string_normalize` RQuickJS intrinsic and a method on `Config` to make it available or unavailable. It defaults to available similar to how we handle other RQuickJS intrinsics.

Also lowers the fuel consumed amount for the empty test because that test was failing for me and increases the threshold to 3% since I can't get the tests to pass on both MacOS and Linux at 2%.

Someone tried to use the `normalize` method on strings and it didn't exist.

## Checklist

### DIFF
--- a/crates/cli/tests/integration_test.rs
+++ b/crates/cli/tests/integration_test.rs
@@ -12,7 +12,7 @@ fn test_empty(builder: &mut Builder) -> Result<()> {
     let mut runner = builder.input("empty.js").build()?;
 
     let (_, _, fuel_consumed) = run(&mut runner, vec![]);
-    assert_fuel_consumed_within_threshold(23_048, fuel_consumed);
+    assert_fuel_consumed_within_threshold(22_448, fuel_consumed);
     Ok(())
 }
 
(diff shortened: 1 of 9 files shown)


crates/cli/tests/integration_test.rs
```diff
@@ -402,7 +402,7 @@ fn run_fn(r: &mut Runner, func: &str, stdin: Vec<u8>) -> (Vec<u8>, String, u64)
 fn assert_fuel_consumed_within_threshold(target_fuel: u64, fuel_consumed: u64) {
     let target_fuel = target_fuel as f64;
     let fuel_consumed = fuel_consumed as f64;
-    let threshold = 2.0;
+    let threshold = 3.0;
```
saulecabrera, line 405: It's not totally clear why this change is needed, but perhaps I'm missing something?

crates/javy/src/runtime.rs
```diff
@@ -128,6 +128,10 @@ impl Runtime {
                 unsafe { intrinsic::BignumExt::add_intrinsic(ctx.as_raw()) }
             }
 
+            if intrinsics.contains(JSIntrinsics::STRING_NORMALIZE) {
+                unsafe { intrinsic::StringNormalize::add_intrinsic(ctx.as_raw()) }
```
saulecabrera, line 132: Could you add an entry here? https://github.com/bytecodealliance/javy/blob/main/docs/docs-using-js-api-support.md -- I guess that page needs a general update, so might be worth doing while you're there, if you feel like it.

crates/cli/tests/integration_test.rs
jeffcharles: If I don't make this change, the `test_empty` test will always fail on one of Linux or MacOS because the difference between the fuel counts on the different operating systems is 2.3% apart with this change.

saulecabrera (approved):

PR #1025, opened 2025-10-06
Update SWC and serde

jeffcharles:
## Description of the change

Updates SWC and serde dependencies.

Dependabot tried making the serde updates and those failed due to SWC trying to use serde APIs that are no longer available.

## Checklist

### DIFF
--- a/crates/codegen/src/js.rs
+++ b/crates/codegen/src/js.rs
@@ -8,7 +8,6 @@ use std::{
     fs::File,
     io::{Cursor, Read},
     path::Path,
-    rc::Rc,
 };
 
 use anyhow::{anyhow, bail, Context, Result};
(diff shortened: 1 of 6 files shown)


crates/codegen/src/js.rs
```diff
@@ -148,7 +145,10 @@ impl JS {
 
     fn parse_module(&self) -> Result<Module> {
         let source_map: SourceMap = Default::default();
-        let file = source_map.new_source_file_from(FileName::Anon.into(), self.source_code.clone());
+        // SWC's source file API works with BytesStr's and initializing a
+        // ByteStr requires either a static str or to own the string so we have
+        // to clone here.
+        let file = source_map.new_source_file(FileName::Anon.into(), self.source_code.to_owned());
```
surma, line 151: The other alternative:
Use `String::leak(...)` in the constructor for `JS` (giving you a `&'static str`) and add a `Drop` implementation for `JS` where you do `Box::from_raw()`. The lifetimes are practically tied anyway, aren’t they? That way we can avoid the clone here. I’ll let you decide whether that’s worth doing.

PR #1062, opened 2025-11-06
Modified the docs/docs-using-js-api-support.md file and the docs/docs-using-extending.md file

AyzeLYC:
My pull request isn't of high priority.

In my different modifications i :

- Removed the line |`TextEncoder`|🚧| Partial support, not fully compliant| in docs/docs-using-js-api-support.md ( that line was in the file twice )

- Added a few more words in docs/docs-using-extending.md to make the informations easier to understand for peoples that are not native english speakers


AyzeLYC:
i'll modify the files if you want but it's mostly to make it easier for peoples to read if they've been reading all the docs and the code for a long time

PR #1067, opened 2025-11-10
Add support for WASI p1 plugins to support Rust 1.91.0

### LINKED ISSUE #1052: Rust 1.91.0 breaks initializing the default plugin and breaks the use of the WASI preview 1 APIs for I/O
When updating from Rust 1.90.0 to Rust 1.91.0 we can't initialize the default plugin as part of building the Javy CLI.

The current approach compiles the Javy plugin to target WASI preview 2, then runs the `javy-plugin-processing` crate which extracts a core module from the generated component, then uses Wizer to initialize the plugin. As part of initializing the plugin, there's some logic in the plugin to read a configuration from stdin. Up to Rust 1.90.0, that used an imported `fd_read` function and worked in any environment that linked WASI preview 1 hostcalls. Starting in Rust 1.91.0, it uses `"wasi:cli/stdin@0.2.4" "get-stdin"`. Since we have the linker configured to trap on unknown imports, this causes a trap during plugin processing.

Beyond that, trying to use the stdio calls in a Javy module will also either lead to an instantiation error due to missing hostcalls (or traps if those environments are similarly configured to trap on unknown exceptions).

jeffcharles:
## Description of the change

- Adds support for WASI preview 1 plugins
- Changes the default plugin to target WASI preview 1
- Updates toolchain file to target latest stable version of Rust instead of 1.90.0

Note that I haven't bumped the plugin API version or import namespace because I did that in #1032 and we haven't published a new version `javy-plugin-api` or released a new version of Javy since that was merged.

Fixes #1052.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: fmt fmt-check lint-wasi-targets test-wasi-targets wasi-targets lint-native-targets test-native-targets native-targets test-wpt test clean cli plugin build-test-plugins build-default-plugin
+.PHONY: fmt fmt-check lint-wasi-targets test-wasi-targets wasi-targets lint-native-targets test-native-targets native-targets test-wpt test clean cli plugin build-test-plugins build-default-plugin vet ci
 .DEFAULT_GOAL := cli
 
 # === Format checks ===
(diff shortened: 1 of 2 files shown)


saulecabrera (commented):
Would you also mind adding
            
```
BINDGEN_EXTRA_CLANG_ARGS_wasm32_wasip1 = "";
```
to `flake.nix`?

saulecabrera (approved):

PR #1068, opened 2025-11-10
Simplify makefile

jeffcharles:
## Description of the change

Use explicit filepaths for dependencies for targets in the Makefile.

With Makefiles, for a target, if all files it depends on exist and have older timestamps, then `make` won't run the commands to rebuild those files. This should allow us to simplify the Makefile.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,7 @@ jobs:
         run: make lint-native-targets
 
       - name: Test
-        run: make test-native-targets-ci
+        run: make test-native-targets
 
       - name: WPT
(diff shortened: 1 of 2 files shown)


jeffcharles:
> What's the difference between this approach and leaving it up to cargo to decide whether a rebuild is needed?

The main one is there isn't currently a way to make Cargo aware that the `javy-plugin` crate needs to be compiled and compiled to `wasm32-wasip2` with a release profile for `javy-cli` to compile to a native target. Otherwise they work under the same principle because Cargo's approach was inspired by Make's.

PR #1069, opened 2025-11-10
cli: Remove `wasm-opt` from build dependencies

saulecabrera:
It's not used as part of the build, but in the crate itself.


crates/cli/Cargo.toml
```diff
@@ -20,6 +20,7 @@ serde = { workspace = true, default-features = false }
 serde_json = { workspace = true }
 javy-codegen = { path = "../codegen/", features = ["plugin_internal"] }
 javy-plugin-processing = { path = "../plugin-processing" }
+wasm-opt = { workspace = true }
```
jeffcharles: I think you can also delete this line. It's only used in the `plugin-processing` and `codegen` crates, there are no references in the `cli` crate.

PR #1079, opened 2025-11-18
make: Correctly format `lint-wasi-targets` target

saulecabrera:
The formatting error in the Makefile prevented the lint step from completing and from reporting the lint errors declared in the `wasi_p1` plugin api.

In CI, this looked like:

  ```
 --target=wasm32-wasip1 --all-targets --all-features -- -D warnings
   /bin/sh: 1: --target=wasm32-wasip1: not found
   make: [Makefile:13: lint-wasi-targets] Error 127 (ignored)

```
Ref: https://github.com/bytecodealliance/javy/actions/runs/19335192688/job/55308089241#step:4:21

Everything worked, given that the non-zero exit is ignored. I'm currently not sure what's the best way to ensure that this doesn't happen in the future.

Additionally, to make targets easier to read, this commit also divides the p1/p2 targets, which will hopefully make this process less error-prone.

### DIFF
--- a/crates/plugin-api/src/wasi_p1.rs
+++ b/crates/plugin-api/src/wasi_p1.rs
@@ -44,32 +44,36 @@ unsafe extern "C" fn cabi_realloc(
     let new_mem = match new_size {
         0 => ZERO_SIZE_ALLOCATION_PTR,
         // this call to `alloc` is safe since `new_size` must be > 0
-        _ => alloc::alloc(Layout::from_size_align(new_size, alignment).unwrap()),
+        _ => unsafe { alloc::alloc(Layout::from_size_align(new_size, alignment).unwrap()) },
     };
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/plugin-api/src/wasi_p1.rs [hunk 1]
```diff
@@ -44,32 +44,36 @@ unsafe extern "C" fn cabi_realloc(
     let new_mem = match new_size {
         0 => ZERO_SIZE_ALLOCATION_PTR,
         // this call to `alloc` is safe since `new_size` must be > 0
-        _ => alloc::alloc(Layout::from_size_align(new_size, alignment).unwrap()),
+        _ => unsafe { alloc::alloc(Layout::from_size_align(new_size, alignment).unwrap()) },
```
jeffcharles: Should we remove `unsafe`s on the functions if we're adding it in the implementation?

PR #1100, opened 2025-12-03
Use rquickjs-serde implementation

Sytten:
## Description of the change

Use the https://github.com/rquickjs/rquickjs-serde 

This is just a proposal to see if the test passes. If so it would be nice to centralize around a shared implementation.
Happy to give you guys access to the repository to publish new versions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2389,9 +2389,9 @@ dependencies = [
 
 [[package]]
 name = "rquickjs-serde"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Sytten:
@jeffcharles All test are passing now, the issues were:
- BigInt default to String serialization (fixed with a new `strict` mode in rquickjs-serde)
- Error string comparison not having the JSError prefix (fixed by adding the prefix + a comment to not remove it in rquickjs-serde)

saulecabrera (approved):
LGTM, thanks for putting this together. Given Javy's dependency on JSON, before landing this PR would you mind adding us as collaborators on the repo and ideally also in crates.io?

saulecabrera:
Thanks!

PR #1125, opened 2026-02-05
Lock Rust version to 1.92 to workaround WASI p2 bug

### LINKED ISSUE #1115: Build with Rust 1.93 failing
See [this GitHub Action workflow run](https://github.com/bytecodealliance/javy/actions/runs/21252590780/job/61194566961#step:5:730). It's reporting an out of memory error from Rust.

I'm able to reproduce the failure when running locally on MacOS. Downgrading from Rust 1.93.0 to Rust 1.92.0 seems to resolve the issue. It also looks like this only affects WASI preview 2 and not WASI preview 1.

Looks like the failing test is [this one](https://github.com/tc39/test262/blob/249657722525cc8e43b1ddb91f8df0b4b011fcf6/test/built-ins/JSON/parse/15.12.1.1-g4-4.js#L11-L13).

jeffcharles:
## Description of the change

Locks the Rust version to 1.92.0 instead of using stable which is currently 1.93.0.

Fixes #1115 

The 262 tests are broken for Rust 1.93.0 with WASI preview 2 due to a bug in Rust's standard library for WASI preview 2. The bug is fixed when run with a nightly for Rust 1.95.0 so presumably we can switch back to using stable when 1.94 or 1.95 is released on the stable channel.

## Checklist

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,8 @@
 [toolchain]
-channel = "stable"
+# Rust 1.93.0 has a bug where WASI p2 doesn't correctly close file descriptors causing tests to fail.
+# cargo 1.95.0-nightly (fe2f314ae 2026-01-30) runs correctly so presumably we can switch back to stable for 1.94 or
+# 1.95.
+channel = "1.92.0"
 components = ["clippy", "rustfmt"]
 targets = ["wasm32-unknown-unknown", "wasm32-wasip1", "wasm32-wasip2"]
 profile = "default"


saulecabrera:
We've been in this situation a couple of times [[1](https://github.com/bytecodealliance/javy/pull/1053)] [[2](https://github.com/bytecodealliance/javy/pull/205)], I wonder if we should define a supported Rust version policy, which would hopefully avoid intermittent/unexpected failures like this one. I'm fully aware that this is happening with stable, but with a clearly defined policy, the Rust version upgrade process (e.g., v1 -> v2) is explicit rather than implicit. Thoughts?

jeffcharles:
I have two reasons I prefer defaulting to stable with an automatic update of Rust versions:
1. Updating our dependencies will sometimes force us to use a newer Rust version. In some cases we need to adopt those dependency updates due to CVEs.
1. We find out about behaviour that breaks in the latest stable Rust version or with Clippy much sooner so that gives us more time to update our code if we need to.

The case I don't want to run into in the future if we lock our Rust version is: say we have a dependency that has a CVE filed against it so we have to update it, but the updated dependency requires a newer Rust version than we're using. So we try to update our Rust version, but then we get CI failures from updating Rust, so then we have to do other code changes that might take a while (for example, what happened with WASI p2 removing the WASI p1 imports for stdio operations), before we can ship a change to fix the CVE. But if we kept up with the latest stable Rust, we find the problem, we can file an issue to fix CI, we can lock to the last working stable version, then perform the CVE update, then do whatever work we need to do to fix CI with the latest stable Rust, then switch to latest stable. It's more PRs but we find problems sooner so we have the option of fixing the problems sooner.

PR #1158, opened 2026-03-26
Fix JS runtime option help

### LINKED ISSUE #1157: javy build -J help errors, does not list "intrinsic options"
Operating system: Linux
Processor architecture: x86_64
Rust version: rustc 1.96.0-nightly (03749d625 2026-03-14)
Javy version: 8.1.0

## Problem
```
javy build --help
Generates WebAssembly from a JavaScript source

Usage: javy build [OPTIONS] <INPUT>

Arguments:
  <INPUT>  Path of the JavaScript input file

Options:
  -o <OUTPUT>              Desired path of the WebAssembly output file [default: index.wasm]
  -C, --codegen <CODEGEN>  Code generation options. Use `-C help` for more details
  -J, --javascript <JS>    JavaScript runtime options. Use `-J help` for more details
  -h, --help               Print help
```

## Error Message and Context
```
javy build -J help
error: the following required arguments were not provided:
  <INPUT>

Usage: javy build --javascript <JS> <INPUT>

For more information, try '--help'.
```

Reference: https://github.com/bytecodealliance/javy/issues/702

Compare `javy build -C help` output:

```
javy build -C help
Available options for codegen

-C  dynamic[=y|n]
 Creates a smaller module that requires a dynamically linked QuickJS plugin Wasm module to execute (see `emit-plugin` command).

-C  wit=path
 Optional path to WIT file describing exported functions. Only supports function exports with no arguments and no return values.

-C  wit-world=val
 Optional WIT world name for WIT file. Must be specified if WIT is file path is specified.

-C  source[=omitted|compressed|uncompressed]
 How to embed the JavaScript source in a custom section in the generated WebAssembly module. Options are `omitted`, `compressed`, and `uncompressed`. `compressed` enables source code compression which generates smaller WebAssembly files at the cost of increased compile time.

-C  plugin=path
 Optional path to Javy plugin Wasm module. Required for dynamically linked modules. JavaScript config options are also not supported when using this parameter.

-C  deterministic[=y|n]
 Produce deterministic output by using fixed clocks and constant zero-filled RNG during pre-initialization. Ensures identical input always produces identical output. Security note: both secure_random and insecure_random become non-secure.
```

jeffcharles:
## Description of the change

Move handling `-J help` into the options parser instead of performing it later. This means we can inline `JsGroupOption` wherever `JsGroupValue` was used since we're not propagating that help was requested. Also update the error message emitted when someone tries to use `-J` with a plugin to mention it's not supported when specifying a plugin.

Fixes #1157.

I opted to move the handling with `-J help` into the parser because it seems like making `input` optional in this one particular case would introduce more complexity compared to just loading the default plugin to provide the runtime options. I updated the error message for using `-J` with plugins because it could be confusing for a user if the `-J help` message says a particular option is supported and then they get an error message saying that option is not supported when trying to use it with a plugin.

## Checklist

### DIFF
--- a/crates/cli/tests/integration_test.rs
+++ b/crates/cli/tests/integration_test.rs
@@ -431,6 +431,24 @@ fn test_js_help() -> Result<()> {
 }
 
 #[test]
+fn test_codegen_help() -> Result<()> {
+    let output = Command::new(env!("CARGO_BIN_EXE_javy"))
+        .args(["build", "-C", "help"])
+        .output()?;
+    assert!(
+        output.status.success(),
+        "build -C help failed: {}",
+        str::from_utf8(&output.stderr)?
+    );
+    let stdout = str::from_utf8(&output.stdout)?;
+    assert!(
+        stdout.contains("Available options for codegen"),
+        "unexpected help output: {stdout}"
+    );
+    Ok(())
+}
+
+#[test]
 fn test_init_plugin() -> Result<()> {
     let engine = Engine::default();
     let mut linker = Linker::new(&engine);


crates/cli/tests/integration_test.rs
```diff
@@ -414,6 +412,24 @@ fn test_source_code_omitted(builder: &mut Builder) -> Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_js_help() -> Result<()> {
```
saulecabrera: Skimming this file it seems that we don't have tests for `build -C help`? Could you also add a test for that as well?

Additionally, could we add an integration test that exercises that error cases (i.e., plugins)

saulecabrera (approved):

PR #1167, opened 2026-04-03
Coherently handle the `publish` and the `version` entries in each package

saulecabrera:
This is the first of a series of patches to improve how we handle version updates and releases. Note that my intention is not change our current mechanism, but to improve it so that there are less manual steps along the way, ideally making it less error prone.

This commit set the `publish` key in each package to `false` and sets the version to `0.0.0`, to explicitly the mark the package as not publishable, as documented in
https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field.

The main motivation for this change is to prevent certain packages from being published by mistake, either manually or in light of a future publishing mechanism.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1704,7 +1704,7 @@ dependencies = [
 
 [[package]]
 name = "javy-plugin"
-version = "0.2.0"
+version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


jeffcharles (approved):
Thank you for adding these!

PR #1169, opened 2026-04-03
Introduce the `javy-release` crate

saulecabrera:
This commit introduces the `javy-release` crate.

The idea behind this crate is to provide a utility to partially automate our release process, at least initially. In its current form, this crate adds the necessary functionality to toggle the release/dev versions according to our versioning system.

Even though publishing to crates.io should be possible, I've opted exclude such functionality in the initial state of this crate mostly for simplicity. Said functionality can be added as a follow-up if this crate gets adopted.

Furthermore, the principal motivation for this change is to ease the release process, which is currently manual and potentially error prone. I've gone through this process twice, and in every instance it's unclear to me the order of operations that I need to follow to start a successful release.

### DIFF
--- a/release/src/lib.rs
+++ b/release/src/lib.rs
@@ -53,10 +53,13 @@ impl PublishableCrate {
     }
 
     fn update_changelog(&self, version: &str, date: &str) -> Result<()> {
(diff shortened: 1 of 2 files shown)


jeffcharles (approved):
I'm all for more automation

PR #1186, opened 2026-04-26
Remove orphan temporary files; support Deno and Bun

### LINKED ISSUE #1185: Node.js example leaves orphan files in temporary directory; support Deno and Bun in example
Operating system: Linux
Processor architecture: x86_64
Rust version: rustc 1.96.0-nightly (03749d625 2026-03-14)
Javy version: javy 8.1.0

## Problem
Those temporary files created in the Node.js example never get removed. 

Deno, Bun, txiki.js, and other JavaScript runtimes that support or can support WASI Preview 1 are not incorporated into the example.

## Error Message and Context

Observer files piling up in `/tmp` on Linux.

No support for JavaScript runtimes other than Node.js `node:wasi` module.

## Prospective solution

Update and modify the Node.js embedding example to support more than just Node.js JavaScript runtimes, starting with Deno, and Bun, then txiki.js, et al.

I have written such a modified example.

guest271314:
Fixes https://github.com/bytecodealliance/javy/issues/1185

## Description of the change

Remove orphan temporary files; support Deno and Bun

The code in the current example results in orphan files being left in temporary directory. Support Deno and Bun using the same code. Eventually support more JavaScript runtimes, such as txiki.js.

## Checklist

- [ ] I've updated the default plugin import namespace and incremented the major version of `javy-plugin-api` if the QuickJS bytecode has changed.
- [ ] I've updated the relevant CHANGELOG files if necessary. Changes to `javy-cli`, `javy-plugin`, and `javy-plugin-processing` do not require updating CHANGELOG files.

### DIFF
--- a/docs/docs-using-nodejs.md
+++ b/docs/docs-using-nodejs.md
@@ -108,14 +108,7 @@ function writeOutput(output) {
 
 `host.js`
 ```javascript
-import { readFile } from "node:fs/promises";
-import {
-  closeSync,
-  openSync,
-  readFileSync,
-  rmSync,
-  writeFileSync,
-} from "node:fs";
+import { open, readFile, rm, writeFile } from "node:fs/promises";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 import { randomUUID } from "node:crypto";
@@ -147,22 +140,24 @@ async function runJavy(pluginModule, embeddedModule, input) {
   const stderrFilePath = join(workDir, `stderr.wasm.${uniqueId}.txt`);
 
   // 👋 send data to the Wasm instance
-  writeFileSync(stdinFilePath, JSON.stringify(input), { encoding: "utf8" });
+  await writeFile(stdinFilePath, JSON.stringify(input), { encoding: "utf8" });
 
-  const [stdinFileFd, stdoutFileFd, stderrFileFd] = [
-    openSync(stdinFilePath, "r"),
-    openSync(stdoutFilePath, "a"),
-    openSync(stderrFilePath, "a"),
-  ];
+  const [stdinFileFd, stdoutFileFd, stderrFileFd] = await Promise.all([
+    open(stdinFilePath, "r"),
+    open(stdoutFilePath, "a"),
+    open(stderrFilePath, "a"),
+  ]);
+
+  // console.log(stdinFileFd.fd, stdoutFileFd.fd, stderrFileFd.fd);
 
   const wasiOptions = {
     version: "preview1",
     returnOnExit: true,
     args: [],
     env: {},
-    stdin: stdinFileFd,
-    stdout: stdoutFileFd,
-    stderr: stderrFileFd,
+    stdin: stdinFileFd.fd,
+    stdout: stdoutFileFd.fd,
+    stderr: stderrFileFd.fd,
   };
 
   try {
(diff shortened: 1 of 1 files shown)


jeffcharles (changes requested):
I appreciate this took some time to put together but this is way too much code to review and maintain for a docs page. I also want to keep the JS host example code simple so it's easy for a reader to understand, it's not intended to be production-ready code. I'm okay with changes to broaden support for JS runtimes that have existing WASI support and okay with changes to clean up the temp files. This is a more minor point but I'd also prefer if we kept using the async APIs and not the sync ones.

guest271314:
> I appreciate this took some time to put together but this is way too much code to review and maintain for a docs page. 

I will maintain the code. 

Documentation should be correct.

> I also want to keep the JS host example code simple so it's easy for a reader to understand, it's not intended to be production-ready code. 

If somebody/anybody winds up here in Javy world, chances are it ain't they're first rodeo. JavaScript to WebAssembly toolchain is non-trivial. This is one of, if not the only library doing that. I expect folks running `javy` are apt in at least JavaScript, if not Rust and WebAssembly tooling, too.

> I'm okay with changes to broaden support for JS runtimes that have existing WASI support and okay with changes to clean up the temp files. This is a more minor point but I'd also prefer if we kept using the async APIs and not the sync ones.

I can change the sync methods to async methods. 

Basically all JavaScript runtims support WASI. It just takes somebody to write up the code - because every single WASI implementation I've seen is different - it's host defined; there's no canonical coe or implementation.

guest271314:
@jeffcharles  

> but this is way too much code to review and maintain for a docs page.

I think the same general analysis and observation must also apply to `node:wasi`. Except we can't really get at and modify that code without going through `uvwasi`, which in reality winds up being even more code. So, 6 of one, half-dozen of the other.

guest271314:
> but this is way too much code to review and maintain for a docs page.



@jeffcharles I removed all of the exports Javy doesn't expect from WASIP1 host implementations. Only the imports Javy uses (throws when not callable) are included.

jeffcharles:
As I mentioned in my prior feedback, please remove `wasi.js` from this pull request. This project is not the place to define a host runtime agnostic WASI implementation. There is value in providing that implementation, but please provide it elsewhere.

guest271314:
So, just add `rm()` to get rid of those orphan files, and leave everything else, as is?

guest271314:
@jeffcharles 

> As I mentioned in my prior feedback, please remove wasi.js from this pull request. This project is not the place to define a host runtime agnostic WASI implementation. 

Then where would that be? Parent WebAssembly/WASI repository?

Right now, since there is no canonical implementation, using WASI at all in JavaScript is all third-party hit and miss, and maintainer _preference_.

jeffcharles:
Maybe ECMA's TC39 or TC55?

guest271314:
Banned from both. ECMA-262 doesn't specifiy I/O at all. Neither does the Winter CG folks. I brought up the lack of STDIO in JavaScript. I guess they are busy with other stuff. I think WebAssembly/WASI should spell it out https://github.com/WebAssembly/WASI/issues/914. It's their idea. Deno, Node, txiki.js, and other JavaScript runtimes _could_ support this, coming from the horse's mouth. Then there wouldn't be the Node.js and reluctance for others' gear - even though that gear works, folks evidently would rather have no functionality for runtimes other than `node` without some kinda branding/label of an organization. It'd be canonical coming from WebAssembly/WASI. Otherwsie, it will just continue to be arbitrary; from runtime to runtime, library to library; micking the state of I/O in JavaScript right now.

PR #1203, opened 2026-05-05
build: Tighten crate selection

saulecabrera:
The exclude list in the Makefile is arguably unwieldy and makes the process of adding new crates harder than it has to be. This change aims to ease the per-target crate selection for testing and linting though a small function and script that reads each crate's metadata and decides which target it should be built for.

The resulting invocation is very similar to what we had before, e.g. the invocation to lint native targets, with this change looks like the following:

    CARGO_PROFILE_RELEASE_LTO=off cargo clippy -p javy-cli -p javy-codegen -p javy-plugin-processing -p javy-runner -p javy-test-macros -p javy-fuzz -p javy-release \
    --release --all-targets --all-features -- -D warnings

By default all crates are treated as _build on native_ unless stated otherwise. With this approach, the only pre-requisite to add new crates shifts from:

* Adding a new crate
* Decide how and where it should be added to the exclude lists in the Makefile

To:

* Add a new crate with the relevant metadata.

One thing to note is that this change adds overhead to each `make` invocation, since now parsing of the metadata is needed. Basic measurements show that all these invocations take around 240ms, which I believe is fine, given that compilation takes way longer. As a future optimization we can try to cache the invoation initially, and apply filtering on top of the cached result.

This change is motivated by the introduction of a profiler for Javy, which will require introducing several new crates.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,45 @@
 .PHONY: fmt fmt-check lint-wasi-targets lint-wasip1-targets lint-wasip2-targets \
 	test-wasi-targets test-wasip1-targets test-wasip2-targets wasi-targets \
-	lint-native-targets test-native-targets test-native-targets-ci native-targets \
-	test-wpt test-wpt-ci test-all clean cli build-default-plugin build-test-plugins \
-	ci
+	lint-native-targets lint-native-targets-ci test-native-targets \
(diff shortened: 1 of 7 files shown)


jeffcharles (approved):
This is much nicer!

Issue #1206, opened 2026-05-12
Add support for profiling execution of JS on Wasm


saulecabrera:
This issue is meant to track the implementation progress of an _experimental_ profiler for Javy.

#### High-level goals

* Allow developers to understand the behavior of JS programs running on Wasm.
* Output traces and an estimated instruction cost per JS opcode, per JS program
* Allow Javy developers to understand the behavior of JS programs running on Wasm to derive future optimizations.

This work is the materialization of the research work done in collaboration with the Carnegie Mellon WebAssembly Research Center around execution and optimization of dynamic languages on Wasm.

#### Work to be done

- [x] Add a new `javy-profiler` crate, which will be called from CLI
- [x] Add a new `javy-profiler-lib` crate, containing the necessary heuristics to perform bytecode rewriting and instrumentation. This crate will be compiled to `wasm32-wasip1` and will be used by [whamm](https://github.com/ejrgilbert/whamm) as a user-defined library.
- [ ] Upstream the necessary crates to parse and transform QuickJS bytecode
- [ ] Upstream the crates to analyze execution traces
- [ ] Add new commands to the CLI (behind a feature flag) to enable profiling

#### Using the profiler from the CLI

First, build the WebAssemlby, optionally providing a plugin:

```sh
javy build index.js -o index.wasm
```
Next, use the `profile inject` subcommand:

```sh
javy profile inject index.wasm -o index.injected.wasm
```

Next, run the injected Wasm through Wasmtime or your custom
embedding and retrieve the profile results.

Finally, run the analysis command to retrieve the results:

```sh
javy profile analyze index.analysis.bin
```

Optionally we can have a single flow that injects and performs the
analysis in a single step when the default plugin is used.

```
javy profile run index.js
```

saulecabrera:
Worth noting that:

* Profiling will only work with statically linked modules. In order to perform profiling, access to the interpreter loop is needed anyway, plus there is nothing of interest worth profiling in the dynamically linked case aside from a couple of function calls that are not JS specific.
* It is still an open question the format to be used for the profile results.JSON/CSV are a possibility, but this could change since it would also be beneficial to embed the JS source in the analysis output, to ensure that the `analyze` step has all the necessary information.

PR #1208, opened 2026-05-12
Introduce the javy-profiler-lib crate

saulecabrera:
Part of https://github.com/bytecodealliance/javy/issues/1206

This is the first change towards a Wasm-native profiler for Javy generated modules.

Particularly, this change introduces the profiler library crate, which will be used as a user-provided library by
[Whamm](https://github.com/ejrgilbert/whamm).

The library will be compiled to wasm32-wasip1 and will have the following responsibilities:

* Detecting the Wasm function in the target app holding the interpreter dispatch loop
* Detecting the byte-load provenance used to dispatch to the next opcode (index argument to `br_table`)
* Holing the necessary state to construct the JS execution trace
* Generating the execution trace output for analysis

This change includes, point 1 and 2. The remaining functionality will be done in separate pull requests.

Note that, as stated in the issue above, this work is deemed experimental for the time being and this crate will be included through a feature flag when added to the CLI.


crates/profiler-lib/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+mod ai;
+mod state;
+
+use state::State;
+use std::sync::LazyLock;
+
+// TODO: Passing empty bytes is temporary. Whamm currently does not
```
saulecabrera, line 7: I am currently working on adding support for this upstream.

saulecabrera:
Hmm, taking a look at the clippy failures, I was not able to repro locally.

crates/profiler-lib/src/ai.rs
jeffcharles, line 1: Nit: I'd prefer if we called this module `abstract_interpreter` or something similar to disambiguate it.

jeffcharles (approved):
Can you add some docs somewhere about how this will integrate with Whamm and other pieces? I understand if it makes more sense to add that later as part of a future PR.

crates/profiler-lib/src/ai.rs
saulecabrera: I was hoping to raise some VC money with this name. Fixed in https://github.com/bytecodealliance/javy/pull/1208/commits/08ae19691e8af126564006fba001983cd8715424

saulecabrera:
> Can you add some docs somewhere about how this will integrate with Whamm and other pieces? I understand if it makes more sense to add that later as part of a future PR.

I'll add them once I work on the `javy-profiler` crate, which will add the integration with Whamm.

Issue #1210, opened 2026-05-20
Switch from NPM CLI to pnpm


jeffcharles:
## What is the idea?

Migrate off of the `npm` CLI and on to the `pnpm` CLI.

## What problem does it solve?

There has been an increasing number of supply chain attacks involving the `npm` CLI executing build scripts by default. `pnpm` does not execute these scripts by default so it's safer.

saulecabrera:
Fixed in https://github.com/bytecodealliance/javy/pull/1211

PR #1211, opened 2026-05-20
Migrate from npm CLI to pnpm

saga-dasgupta:
## Summary

- Switch the `javy` npm package and the WPT harness from the `npm` CLI to `pnpm`, replacing both `package-lock.json` files with `pnpm-lock.yaml`.
- Pin `pnpm@10.28.0` via the `packageManager` field so local dev and CI use the same version, and add `pnpm/action-setup@v4` to the three workflows that touched npm.
- Update the `Makefile` `test-wpt-ci` target and contributor READMEs to use `pnpm` invocations.

The package continues to publish to the npm registry as `javy`; only the CLI used to install and run scripts changes. Dependabot's `package-ecosystem: npm` entry stays — it reads `pnpm-lock.yaml` natively.

## Motivation

Improve supply-chain security for Javy's JS tooling. `pnpm` mitigates several classes of npm supply-chain risk: it refuses to execute install scripts from transitive dependencies by default, uses a content-addressed store that detects tampered tarballs across projects, and enforces strict dependency isolation so packages can only resolve what they explicitly declared. Pinning `packageManager` makes the toolchain version part of the lockfile contract so contributors and CI all run the same `pnpm`.

Tracks bytecodealliance/javy#1210.


saulecabrera (approved):
Thanks!

PR #1213, opened 2026-06-02
Add the initial implementation of javy-profiler

saulecabrera:
This patch is a follow-up to
https://github.com/bytecodealliance/javy/pull/1208.

It marks the initial work to introduce the profiling functionality as described in https://github.com/bytecodealliance/javy/issues/1206.

Concretely, this patch focuses on the Whamm integration:

* Making the necessary changes to the build system to enable building the profiler library that will be consumed by Whamm.
* Adding a new `profile` command and `profile inject` subcommad to the CLI
* Gating this functionality under a feature flag.

Structure wise I tried to make this change as self-contained as possible (see the changes the makefile, and the introduction of the `profiler` feature flag), allowing for an straightforward removal from tree if we ever decide that this feture should not stabilized.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,8 @@
 	test-wasi-targets test-wasip1-targets test-wasip2-targets wasi-targets \
 	lint-native-targets lint-native-targets-ci test-native-targets \
 	test-native-targets-ci native-targets test-wpt test-wpt-ci test-all \
-	clean cli build-default-plugin build-test-plugins ci
+	clean cli build-default-plugin build-requested-feature-assets \
+	build-all-feature-assets build-test-plugins ci
 .DEFAULT_GOAL := cli
 
 # Selection for linting / testing.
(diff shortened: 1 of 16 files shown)


saulecabrera:
Investigating the failures, unsure if they are fully related to the changes here, I am seeing:

```
Error: Unable to process file command 'output' successfully.
Error: Invalid format '45.0.0'
```

https://github.com/bytecodealliance/javy/actions/runs/26829367024/job/79106160369?pr=1213

crates/profiler/build.rs
saulecabrera, line 1: Note that I could move all this to the CLI, but in the spirit of having this feature self-contained, I decided to put it here, at the cost of some code duplication.

Makefile
```diff
@@ -41,6 +42,33 @@ WASIP1_TEST_CRATES := $(call select_crates_for,wasip1,test)
...
+# https://rust-lang.github.io/rfcs/3028-cargo-binary-dependencies.html).
+# This function queries through the CLI metadata, the command(s)
+# needed to build asset dependencies belonging to any of the specified
+# features, e.g.,
+#
+#     make cli features=profiler
+#
+# it will retrive `cargo build -p javy-profiler-lib
+# --target=wasm32-wasip1 --release` as defined the CLI's Cargo.toml
+define feature_asset_cmds
```
saulecabrera, line 57: An alternative to this would be to define a new target e.g., `make cli-profiler`. I didn't go with that approach since this one seems to be more extensible for other optional upcoming features, e.g., the compiler.

crates/profiler/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+# We do not want to publish this crate. It is only meant to be used by
+# the CLI.
+publish = false
+version = "0.0.0"
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+whamm = { git = "https://github.com/saulecabrera/whamm", branch = "binary-entrypoints" }
```
saulecabrera, line 12: This is temporary until https://github.com/ejrgilbert/whamm/pull/326, is merged. I think this is acceptable, given the experimental nature of this feature. I'm planning to update once the PR is merged.

saulecabrera:
> Investigating the failures, unsure if they are fully related to the changes here, I am seeing:
> 
> ```
> Error: Unable to process file command 'output' successfully.
> Error: Invalid format '45.0.0'
> ```
> 
> https://github.com/bytecodealliance/javy/actions/runs/26829367024/job/79106160369?pr=1213

The error here is that Whamm is pulling in wasmtime v38, but Javy depends on v45. I think that the real fix is to update Whamm to use v45. I'll do that in my branch.

saulecabrera:
> > Investigating the failures, unsure if they are fully related to the changes here, I am seeing:
> > ```
> > Error: Unable to process file command 'output' successfully.
> > Error: Invalid format '45.0.0'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://github.com/bytecodealliance/javy/actions/runs/26829367024/job/79106160369?pr=1213
> 
> The error here is that Whamm is pulling in wasmtime v38, but Javy depends on v45. I think that the real fix is to update Whamm to use v45. I'll do that in my branch.

I ended up fixing the cargo metadata invocation to only account for the wasmtime version of the direct dependencies. Given that this feature is currently experimental, I think this is the right path forward, also, this approach will allow updating wasmtime for the default features while not being blocked by the version used in Whamm

jeffcharles (approved):

PR #1224, opened 2026-06-05
Update to rquickjs 0.12.0

jeffcharles:
## Description of the change

Updates rquickjs to 0.12.0. Also bumps the crate and public versions because of QuickJS bytecode changes and removal of one API. Specifically this version of rquickjs always includes `BigInt` support. The `rquickjs-serde` API also changed so I've updated the serializers. The fuel usage has also dropped across a number of test cases.

It can't be done automatically.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -570,6 +570,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


jeffcharles:
Opened https://github.com/rquickjs/rquickjs-serde/pull/7 to deal with the failing test. After that is addressed, we'll also need to tweak some fuel count assertions since the fuel usage will have dropped.

saulecabrera (approved):

PR #1226, opened 2026-06-11
profiler: Fill in the profiler state functions

saulecabrera:
This patch is a follow-up to https://github.com/bytecodealliance/javy/pull/1213

It fills in the profiler state implementation, so that it can be usable from the whamm monitor.

The most notable change that is not described in the original issue, is the profile result encoding format, described in the `format.rs` module.

### DIFF
--- a/crates/profiler-lib/src/state.rs
+++ b/crates/profiler-lib/src/state.rs
@@ -1,10 +1,11 @@
 //! Instrumentation state to derive probe insertion.
 
 use anyhow::{Result, bail};
-use std::collections::BTreeSet;
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use walrus::ir::{BrTable, Visitor, dfs_in_order};
 use walrus::{FunctionId, FunctionKind, LocalFunction, ModuleConfig};
 
+use crate::format;
 use crate::interpreter;
 
 /// Threshold above which a `br_table` qualifies as the interpreter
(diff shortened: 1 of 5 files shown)


crates/profiler-lib/src/state.rs
```diff
@@ -19,6 +20,11 @@ pub struct State {
     /// Byte offsets of the `i32.load8_u` instructions in the dispatch
     /// function whose values feed the dispatch `br_table`'s index.
     dispatch_loads: BTreeSet<u32>,
+    /// Per function, the byte offsets of the opcodes the
+    /// profiler counts.
+    /// Calcualted eagerly so that at runtime the calculation becomes
```
jeffcharles:
```suggestion
    /// Calculated eagerly so that at runtime the calculation becomes
```

crates/profiler-lib/src/state.rs
```diff
@@ -386,4 +526,94 @@ mod tests {
...
+    }
+
+    #[test]
+    fn exit_func_restores_caller_function() {
+        let mut p = Profiler::new();
+        p.start_func();
+        p.set_func_addr(0xA00);
+        p.set_dispatch_target(1, 0);
+
+	// Nested call.
```
jeffcharles: Nit: This and the comment below are using tabs for alignment while the rest of the code in this file are using spaces.

crates/profiler-lib/src/state.rs
```diff
@@ -386,4 +526,94 @@ mod tests {
...
+
+    #[test]
+    fn set_dispatch_target_attributes_deltas_per_opcode() {
+        let mut p = Profiler::new();
+        p.start_func();
+        p.set_func_addr(0x1000);
+
+        p.set_dispatch_target(5, 0);
+        p.set_dispatch_target(7, 3);
+        p.set_dispatch_target(5, 8)
```
jeffcharles: I don't think this will compile

jeffcharles (commented):
Code mostly looks good. Just a few nits with comments and looks like a semicolon is missing.

crates/profiler-lib/src/state.rs
saulecabrera: Yep, fixed.

crates/profiler-lib/src/state.rs
saulecabrera: I believe this should be fixed.

jeffcharles (approved):

PR #1242, opened 2026-07-27
Add `BytecodeStripping` to `Config` and default it to strip source to reduce bytecode

jeffcharles:
## Description of the change

Adds configurations for `BytecodeStripping` to `Javy::Config` that control whether source code and debug information is included in QuickJS bytecode. Defaults to stripping source but not debug information. Stripping source affects operations like `function.prototype.toString()`. I don't think this should impact profiling. Stripping debug information results in stacktraces no longer including line numbers. Defaulting to stripping source does change Javy's behaviour in that the last few releases implicitly included source code.

I've also added some automated tests to ensure the Wasm modules don't increase in size too much. I'm happy to tweak the percentage threshold.

I have made adding the configuration and changing the default a minor version bump in `javy` and `javy-plugin-api`. Arguably this could be a major version change since the behaviour of the JavaScript could change. However I'm not sure if anyone would actually be relying on source code being included in practice and they have the option of setting the configuration to not strip source code.

There is an open question of whether this would be more appropriate as a codegen configuration instead of a runtime one given this seems more related to codegen than runtime behaviour. However, I think making this a codegen option would involve needing to update the plugin Wasm API to expose an additional parameter for `compile-src` which would be disruptive. I'm not sure if anyone would actually want to configure Javy to not strip the source code in practice. Another option I considered is simply not making it configurable and always stripping the source code. That would be easiest. But perhaps someone does want source code to be included in the bytecode?

We observed that the size of the QuickJS bytecode for the same source code was often substantially larger than with version 7.1.0 of Javy. I've observed it being 30% to 300% larger on real-world test cases I've given it. It looks like this was related to changes to [quickjs-ng where it started to default to including source code in the bytecode](https://github.com/quickjs-ng/quickjs/pull/218).

## Checklist

### DIFF
--- a/crates/javy/src/config.rs
+++ b/crates/javy/src/config.rs
@@ -68,7 +68,7 @@ pub enum BytecodeStripping {
     /// Omit debug information, which also omits source code.
     ///
     /// Error stack traces no longer include line/column locations.
-    Debug,
+    SourceAndDebug,
 }
 
 /// A configuration for [`Runtime`](crate::Runtime).
@@ -296,8 +296,8 @@ mod tests {
         let mut config = Config::default();
         assert_eq!(config.bytecode_stripping, BytecodeStripping::Source);
 
-        config.bytecode_stripping(BytecodeStripping::Debug);
-        assert_eq!(config.bytecode_stripping, BytecodeStripping::Debug);
+        config.bytecode_stripping(BytecodeStripping::SourceAndDebug);
+        assert_eq!(config.bytecode_stripping, BytecodeStripping::SourceAndDebug);
     }
 
     #[cfg(feature = "json")]
--- a/crates/javy/src/config.rs
+++ b/crates/javy/src/config.rs
@@ -49,6 +49,28 @@ bitflags! {
     }
 }
 
+/// Controls which optional information QuickJS retains in compiled bytecode.
+///
+/// Stripping source code reduces bytecode size, but changes
+/// `Function.prototype.toString()` to no longer return the function's source
+/// code. Stripping debug information further reduces bytecode size, but removes
+/// line/column locations from JavaScript error stack traces.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub enum BytecodeStripping {
+    /// Retain source code and debug information.
+    None,
+    /// Omit source code while retaining debug information.
+    ///
+    /// For compiled functions, `Function.prototype.toString()` no longer
+    /// returns the function's source code.
+    #[default]
+    Source,
+    /// Omit debug information, which also omits source code.
+    ///
+    /// Error stack traces no longer include line/column locations.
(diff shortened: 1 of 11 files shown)


saulecabrera (commented):
Looks reasonable to me. One thing that is not clear is why we need to add two flags? Said differently, what is the use of keeping the debug information and not keeping the source? My understanding is that the debug information is directly relevant to the original source that was compiled to bytecode. What I would probably expect here is, having a single flag:

- `debug = true`: keeps both source and debug information
- `debug = false`: strips both

I do not think this affects profiling, but profiling will eventually rely on the source being present, so having both present will probably be a pre-requisite.

jeffcharles:
AFAICT the debug data contains the line and column locations when errors happen including lines used in stacktraces. So you could omit the source but the the debug information would still have useful output for debugging exceptions which I'd like to keep in place. That said, I don't think the reverse is true. Stripping the debug data looks like it strips the source. So perhaps this should be an enum of `Strip::None`, `Strip::Source`, `Strip::Debug` instead of two booleans. That said, I don't have a use case for stripping the debug data, it just seemed a logical thing to add given it's set in the same place as `strip_source`, but I'm also fine with just creating a flag for stripping the source.

crates/javy/src/config.rs
```diff
@@ -49,6 +49,28 @@ bitflags! {
...
+    /// Omit source code while retaining debug information.
+    ///
+    /// For compiled functions, `Function.prototype.toString()` no longer
+    /// returns the function's source code.
+    #[default]
+    Source,
+    /// Omit debug information, which also omits source code.
+    ///
+    /// Error stack traces no longer include line/column locations.
+    Debug,
```
saulecabrera: Looking at the latest main:

```c
 if (s->allow_debug) {
        bc_put_atom(s, b->filename);
        bc_put_leb128(s, b->line_num);
        bc_put_leb128(s, b->col_num);
        bc_put_leb128(s, b->pc2line_len);
        dbuf_put(&s->dbuf, b->pc2line_buf, b->pc2line_len);
        if (s->allow_source && b->source) {
            bc_put_leb128(s, b->source_len);
            dbuf_put(&s->dbuf, b->source, b->source_len);
        } else {
            bc_put_leb128(s, 0);
        }
    }
```

> Stripping the debug data looks like it strips the source

It looks like this is actual behavior, so the state hierarchy is None ⊂ Source ⊂ Debug.

With this in mind, I think that it is not really possible to disable debug and allow source.

I think the enum is the right call though, however, given the above, what do you think of renaming the enum arms to:

* `None`
* `Source`
* `All` (alterntively `SourceAndDebug`)

Mostly so that the enum arm reflects the actual behavior?

jeffcharles: Yeah sure

saulecabrera (approved):

